PR #9, opened 2015-01-25
Improve performance of pairwise! for SqEuclidean using sumabs2(x, 1)

timholy:
This has a huge beneficial impact in low dimensions (more than 10x), but it's not obvious what will happen in high dimensions. Reference: https://groups.google.com/d/msg/julia-users/G4ktbB-n7ug/8iF2WcaeQb8J

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -51,13 +51,17 @@ evaluate{T <: Number}(dist::SqEuclidean, a::T, b::T) = abs2(a-b)
 sqeuclidean{T <: Number}(a::T, b::T) = evaluate(SqEuclidean(), a, b)
 
 function pairwise!(r::AbstractMatrix, dist::SqEuclidean, a::AbstractMatrix, b::AbstractMatrix)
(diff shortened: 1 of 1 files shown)


johnmyleswhite:
LGTM

timholy:
I did some performance testing. After warmup:
```julia
julia> a = rand(10^6, 4);

julia> @time Distances.sumsq_percol(a)
elapsed time: 0.007038322 seconds (480 bytes allocated)
4-element Array{Float64,1}:
 333225.0
 333016.0
 333299.0
 333771.0

julia> @time sumabs2(a, 1)
elapsed time: 0.007787581 seconds (464 bytes allocated)
1x4 Array{Float64,2}:
 333225.0  333016.0  333299.0  333771.0
```
I would say that difference isn't enough to worry about, so I think is is probably fine in general.

johnmyleswhite:
Thanks, Tim.

timholy:
Thanks for reviewing!

PR #17, opened 2015-08-19
Return Nan if some element equals NaN

matthieugomez:
Currently, we have

```julia
x = fill(1.0, 10)
y = fill(NaN, 10)
chebyshev(x, y)
# 0.0
```

With this pull request, it returns NaN.


matthieugomez:
max(0, NaN) = 0 so the current behavior makes sense

Issue #26, opened 2015-10-28
RFC: type -> immutable


KristofferC:
Would it make sense for the types that has internal fields (like `Minkowski`) to be immutables to elide unnecessary loads?

andreasnoack:
Yes. I think so. I think you are better off with just constructing a new instance if you need a different value.

PR #27, opened 2015-10-31
make Minkowski an immutable

KristofferC:


IainNZ:
Would anyone be using this field?

KristofferC:
Ref  #26

Imo, this should not be a `type` because it so cheap to just create a new one. Having it as an `immutable` would make vectorization, load hoisting etc possible.

IainNZ:
Yeah it definitely seems like a natural immutable. Just worried about breaking something in user code. I don't really want to take that responsibilty for a package I don't use though... maybe you should take ownership of the package?

KristofferC:
Yeah, it is breaking, that's right. Looking through some more there are a few more "obvious" immutables. 

Not sure what the best course of actions are. If I get collaborator status I could tag a last v0.2 version, do the immutable fixes and bump to v0.3 or something..

IainNZ:
Sounds sensible
@andreasnoack @johnmyleswhite or someone with admin, can you add @KristofferC to collaborators so he can maintain this package?

PR #29, opened 2015-11-15
unify the result_type method

### LINKED ISSUE #28: Using Float32 matrices in pairwise dispatches to generic matmul.
The reason is that

```
julia> Distances.result_type(Euclidean(), Float32, Float32)
Float64
```

so later we make a matrix multiplication between a Float32 and Float64 matrix which dispatches to the generic version. I'll try fix it.

KristofferC:
Ref #28, fixes #28

Problem was that the `result_type` in `generic` had different signatures than the ones in `metric` so the specialized `result_types` were never used when `pairwise` was used. This unifies the `result_type` signatures.

Benchmark:

```jl
N = 10000;
dim = 25;
data = rand(Float32, dim, N);
query = rand(Float32, dim, N);

# Master
julia> @time pairwise(SqEuclidean(), data, query);
  2.435364 seconds (37 allocations: 763.017 MB, 0.34% gc time)

# PR
julia> @time pairwise(SqEuclidean(), data, query);
  0.222634 seconds (31 allocations: 381.547 MB, 0.88% gc time)
```

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -24,7 +24,7 @@ abstract Metric <: SemiMetric
 
 # Generic functions
 
-result_type(::PreMetric, T1::Type, T2::Type) = Float64
+result_type(::PreMetric, ::AbstractArray, ::AbstractArray) = Float64
 
 
 # Generic column-wise evaluation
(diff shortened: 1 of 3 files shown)


KristofferC:
Will merge in 3-4 days if no one objects.

PR #30, opened 2015-11-17
RFC: immutable ALL the things

### LINKED ISSUE #26: RFC: type -> immutable

KristofferC:
Changes the types that has internal fields to immutables to enable various optimizations. Technically breaking. Fixes #26


andreasnoack:
:+1:

KristofferC:
Since the other Immutable change was merged by lindahua I will take it he is ok with this too.

PR #33, opened 2015-11-24
RFC: move to only support 0.4

### LINKED ISSUE #32: RFC: Remove dependency on ArrayViews and use slice instead.
This could potentially fix #5 since slices works with sparse matrices.

From `ArrayViews.jl`

> Reasons to prefer ArrayViews:

>Construction of SubArrays is frequently (but not always) 2-4 times slower than construction of views. If you are constructing many column views, ArrayViews may still be the better choice.

However, the construction times of the views are not a bottle neck here so there seems to me to be no reason to keep the `ArrayViews` dependency.

KristofferC:
fixes #32

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,4 @@
+*.jl.cov
+*.jl.*.cov
+*.jl.mem
+*.ji
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ os:
     - linux
     - osx
 julia:
-    - 0.3
     - 0.4
     - nightly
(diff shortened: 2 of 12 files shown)


JohnHBrock:
LGTM. Might be worth adding some tests for sparse matrices.

KristofferC:
Rebased on masters and updated tests to use the more recent `BaseTestNext` which is merged into v0.5.

This is good to go from my point of view. Will merge this and tag version v0.4 in METADATA in a few days if no one objects.

PR #42, opened 2016-05-15
Compat update for String rename

StefanKarpinski:


yuyichao:
Test only so should be fine.

PR #44, opened 2016-06-13
fix type instability in Jaccard for float input

KristofferC:


KristofferC:
From https://groups.google.com/forum/#!topic/julia-users/0uIQAs44Org

> We see that using abs is faster.

KristofferC:
Tested it and it is the same for Bools and faster for Floats.

Issue #45, opened 2016-06-14
Would it make sense to move StatsBase's distances/divergences here?


ararslan:
Currently StatsBase defines the L2, squared L2, L1, and L-inf distances, as well as the generalized Kullback-Leibler divergence, and the mean absolute, max absolute, mean squared, and root mean squared deviations. It seems to me that those may be better served in a package dedicated to distance metrics such as this one, particularly since other functions in StatsBase don't appear to rely on them. Thoughts?

ararslan:
Wow, the amount of feedback here is overwhelming. Not all at once, now! 😜

kmsquire:
@ararslan, it's mostly the case that JuliaStats has a dearth of developers, and no one has really taken responsibility for this package recently. 

To me, your proposal makes sense. It would be good to know how many (and which) packages use any of those functions, and to be sure there's a good plan to deprecate them. 

If no one else responds here soon, can you open a separate issue at StatsBase?  Shouldn't really be necessary, since most of the same developers will be reading both, but since that is the core stats package, there might be a stronger response.

nalimilan:
Makes sense to me too. They feel a bit lonely in StatsBase.

ararslan:
I'd be more than happy to help out with JuliaStats stuff if active devs are scarce!

Do you know of a good way to go about determining which packages rely on the StatsBase distance functions? Deprecating them could be done in a similar manner to the prime number functions in base Julia, for which the deprecation message says to use the Primes package. In this case, it could be as simple as s/Primes/Distances/ and figuring out how long we want to keep them deprecated before removing them.

Sure, if no one else responds soon I'll open an issue on StatsBase.

nalimilan:
It's a good idea to deprecate them anyway, as private code can rely on them. The general principle is that deprecations are only removed in the next major version (or before `1.0`, when bumping the minor version number). See `base/deprecated.jl` in Julia Base about that process.

ararslan:
Thanks for the info about the deprecation process @nalimilan, that's really helpful!

KristofferC:
More distances in `Distances.jl` sounds good to me.

richardreeve:
[...]

PR #49, opened 2016-07-15
Add in Rényi divergences

richardreeve:
Added in Rényi divergences, as discussed in #45 and JuliaStats/StatsBase.jl#131. Goes with Rényi entropy pull for JuliaStats/StatsBase.jl#190 repository.


README.md
```diff
@@ -138,6 +139,7 @@ Each distance corresponds to a distance type. The type name and the correspondin
 |  CorrDist            |  corr_dist(x, y)         | cosine_dist(x - mean(x), y - mean(y)) |
 |  ChiSqDist           |  chisq_dist(x, y)        | sum((x - y).^2 / (x + y)) |
 |  KLDivergence        |  kl_divergence(x, y)     | sum(p .* log(p ./ q)) |
+| RenyiDivergence   | renyi_divergence(x, y, k) | log(sum( x .* (x ./ y) .^ (k - 1))) / (k - 1) |
```
ararslan: Can you make the `|`s line up? Not that it actually matters, it's just bugging me. 😛

src/metrics.jl
```diff
@@ -141,6 +163,28 @@ chisq_dist(a::AbstractArray, b::AbstractArray) = evaluate(ChiSqDist(), a, b)
...
+
+@inline eval_op(dist::RenyiDivergence, ai, bi) = ai, (ai == 0 ? zero(ai) :
+                                                      (dist.is_normal ?
+                                                       ai .* ((ai ./ bi) .^ dist.p) :
+                                                       (dist.is_zero ?
+                                                        bi :
+                                                        (dist.is_one ?
+                                                         ai * log(ai / bi) :
+                                                         # otherwise α = ∞
+                                                         ai / bi))))
```
ararslan: This is pretty hard to read. Would you mind putting it into the long form function syntax with `if` statements in place of ternaries? E.g.
```julia
@inline function eval_op(dist::RenyiDivergence, ai, bi)
    if ai == 0
        zero(ai)
    ...
end
```

richardreeve:
@ararslan These are both addressed now, but the checks will all fail because of JuliaStats/StatsBase.jl#192.

richardreeve:
@ararslan Okay, sorry for all of fiddling, but the tests now increase coverage except in places which have nothing to do with me - I don't know why they now show as not covered, but they clearly are when you look at the coveralls transcript (and they were correctly identified before).

ararslan:
Coverage can be a little weird sometimes. Don't sweat a decrease of 0.3%. :wink:

ararslan:
I don't know much about Rényi divergences so I don't think I'm the best person to perform a thorough code review here. The code looks fine to me, but it'd be great if someone here who knows more about this stuff could chime in.

richardreeve:
I appreciate this is only a partial solution, but there is reasonably extensive testing of the code. Otherwise it may be years before a developer who uses Rényi entropies comes along to check it. For what it's worth, the relevant bit of Renyi's paper is attached, though it doesn't deal with the cases alpha == 0 or alpha == Inf.
renyi

richardreeve:
Note that for historical reasons he uses base 2 for logs, whereas we use natural logs (naturally!).

src/metrics.jl
```diff
@@ -141,6 +163,53 @@ chisq_dist(a::AbstractArray, b::AbstractArray) = evaluate(ChiSqDist(), a, b)
...
+    elseif dist.is_zero
+        return ai, bi
+    elseif dist.is_one
+        return ai, ai * log(ai / bi)
+    else # otherwise q = ∞
+        return ai, ai / bi
+    end
+end
+
+@inline function eval_reduce{T<:AbstractFloat}(dist::RenyiDivergence,
```
KristofferC: Isnt this function type unstable? In some cases you return a `::T` and in some caes `(::T, ::T)`?

PR #63, opened 2017-02-07
Add a precision threshold to Euclidean

timholy:
On master we have this:
```julia
julia> using Distances

julia> X = [0.1 0.2; 0.3 0.4; -0.1 -0.1]
3×2 Array{Float64,2}:
  0.1   0.2
  0.3   0.4
 -0.1  -0.1

julia> pd = pairwise(Euclidean(), X, X)
2×2 Array{Float64,2}:
 7.45058e-9  0.141421
 0.141421    0.0     
```
Obviously the diagonal entries of `pd` should be 0. Here it's not because of the strategy (chosen for efficiency) to calculate dot products and then compute the square-distance by `sa + sb - 2*r`. This PR recalculates the distance by `\sum (ai-bi)^2` when the gap between the points is below some threshold. It should be non-breaking, because the fallback threshold is 0.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -6,8 +6,12 @@
 #
 ###########################################################
 
-type Euclidean <: Metric end
-type SqEuclidean <: SemiMetric end
+immutable Euclidean <: Metric
+    thresh::Float64
+end
+immutable SqEuclidean <: SemiMetric
+    thresh::Float64
+end
 type Chebyshev <: Metric end
 type Cityblock <: Metric end
 type Jaccard <: Metric end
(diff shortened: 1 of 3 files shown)


src/metrics.jl
```diff
@@ -289,6 +293,10 @@ end
 end
 rogerstanimoto{T <: Bool}(a::AbstractArray{T}, b::AbstractArray{T}) = evaluate(RogersTanimoto(), a, b)
 
+
+# Utility
+sqr(x) = x*x
```
ararslan: Is this necessary? It seems you could just do `^2` as needed in the uses below.

timholy: I swear I've seen cases (recently) where `x^2` calls `Base.power_by_squaring` (which is a performance disaster), but I'm having trouble coming up with a concrete example. I can change it.

ararslan: Yeah, I'm not sure under what circumstances it falls back to `power_by_squaring`. I'd say go with the simpler `^2` for now and we can revisit it if we find that things are slooooow, if that seems reasonable.

timholy:
Oh, BTW I've tested performance and don't see any negative impact.

timholy:
What are your thoughts on the following statement? "We added this to `Euclidean`, so we have to add it to `SqEuclidean`." For consistency it seems desirable. But in d9956cd6, I notice that some bozo (`git blame` says the name is "timholy", whoever he is) split out `pdist!` as a separate function. It's not exported, but the `pdist!` interface doesn't allow one to make this kind of correction, since the original coordinates are not provided. The commit message suggests a purpose other than to provide a separate "hook" into the function. So deleting `pdist!` seems likely to be OK. (Or we could keep it but not use it.)

KristofferC:
If it is unexported and no longer needed,  deleting seems the best.

KristofferC:
Would be nice to have a word or two about this in the README.

timholy:
Done.

KristofferC:
LGTM, is this good to go? Travis fails on nightly due to fixed macro hygiene.

timholy:
From my perspective, yes. (Someone else has to hit merge, though.)

timholy:
Tag?

KristofferC:
After https://github.com/JuliaStats/Distances.jl/pull/64

PR #64, opened 2017-02-07
update for 0.6

KristofferC:


KristofferC:
I'm gonna drop 0.4 support since for the `.` function calls.

src/generic.jl
```diff
@@ -117,18 +117,12 @@ end
 function pairwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix)
     m = size(a, 2)
     n = size(b, 2)
-    r = Array(result_type(metric, a, b), (m, n))
+    r = Matrix{result_type(metric, a, b)}(m, n)
     pairwise!(r, metric, a, b)
 end
 
-function pairwise(metric::PreMetric, a::AbstractMatrix)
+function pairwise(metric::Union{PreMetric, SemiMetric}, a::AbstractMatrix)
```
timholy:
```julia
julia> supertype(SemiMetric)
Distances.PreMetric                                                                                                                    
```

test/REQUIRE
```diff
@@ -1 +1 @@
-BaseTestNext
+julia 0.5
```
timholy: wrong REQUIRE file

KristofferC: woops!

timholy:
Really nice work, @KristofferC.

KristofferC:
Thanks, good to go from my point of view? Mac workers are sleeping as usual. Any more comments?

timholy:
LGTM.

KristofferC:
I wonder if attobot is activated here.. I'll try

KristofferC:
Yep, cool: https://github.com/JuliaLang/METADATA.jl/pull/7864

PR #58, opened 2016-10-27
Add distances, divergences, and deviations from StatsBase

ararslan:
Fixes #45.

Request for comment! How does it look? I'd like to hear from anyone, but especially @KristofferC since he seems to have contributed the most to this package.

PR #67, opened 2017-03-28
aa/statsbase migration fix

richardreeve:
@ararslan I'm not sure if this is the right way to correct a pull request - I haven't tried before - but I believe these are the fixes you need for #58 - both to correct NormRMSDeviation to be a PreMetric, and to add CVRMSDeviation.


ararslan:
Yep, making a pull request against my branch is exactly the right way to do it. Thanks for your help!

PR #68, opened 2017-03-31
Tests for correct Pre/Semi/Metric subtyping

richardreeve:
I've noticed while working on #58 that there is a fairly ad hoc testing of whether the subtyping is correct for the distance measures defined here - there is much testing of equality to zero for self-distances, but not symmetry (for `SemiMetric`s) and the triangle inequality (for `Metric`s). I've introduced a test `@test_metricity`, that checks what each distance is a subtype of and tests the appropriate conditions.

There is now a systematic test for each distance measure to check it does not overclaim it's "metricity" (it can't check for underclaiming, but that is a less serious problem). This exposed a series of problems to do with floating point rounding errors for `CorrDist`, `CosineDist` and `RenyiDivergence`. This kind of issue has been identified before (e.g. #13), which I have now fixed (for `RenyiDivergence`), or relaxed the zero equality constraints (for the others).

I then updated the benchmarks and cleaned up a lot of the other testing code to remove unnecessary tests and move to a standard Julia style as seen in the rest of the package (e.g. 1. -> 1.0). Is this useful to people?

### DIFF
--- a/README.md
+++ b/README.md
@@ -185,7 +185,7 @@ julia> pairwise(Euclidean(1e-12), x, x)
 
 The implementation has been carefully optimized based on benchmarks. The Julia scripts ``test/bench_colwise.jl`` and ``test/bench_pairwise.jl`` run the benchmarks on a variety of distances, respectively under column-wise and pairwise settings.
 
-Here are benchmarks obtained on Linux with Intel Core i7-4770K 3.5 GHz.
+Here are benchmarks obtained running Julia 0.5.1 on a late-2016 MacBook Pro running MacOS 10.12.3 with an quad-core Intel Core i7 processor @ 2.9 GHz.
 
 #### Column-wise benchmark
 
@@ -193,28 +193,32 @@ The table below compares the performance (measured in terms of average elapsed t
 
 |  distance  |  loop  |  colwise  |  gain  |
 |----------- | -------| ----------| -------|
-| SqEuclidean | 0.012308s |  0.003860s |  3.1884 |
-| Euclidean | 0.012484s |  0.003995s |  3.1246 |
-| Cityblock | 0.012463s |  0.003927s |  3.1735 |
-| Chebyshev | 0.014897s |  0.005898s |  2.5258 |
-| Minkowski | 0.028154s |  0.017812s |  1.5806 |
-| Hamming | 0.012200s |  0.003896s |  3.1317 |
-| CosineDist | 0.013816s |  0.004670s |  2.9583 |
-| CorrDist | 0.023349s |  0.016626s |  1.4044 |
-| ChiSqDist | 0.015375s |  0.004788s |  3.2109 |
-| KLDivergence | 0.044360s |  0.036123s |  1.2280 |
-| JSDivergence | 0.098587s |  0.085595s |  1.1518 |
-| BhattacharyyaDist | 0.023103s |  0.013002s |  1.7769 |
-| HellingerDist | 0.023329s |  0.012555s |  1.8581 |
-| WeightedSqEuclidean | 0.012136s |  0.003758s |  3.2296 |
-| WeightedEuclidean | 0.012307s |  0.003789s |  3.2482 |
-| WeightedCityblock | 0.012287s |  0.003923s |  3.1321 |
-| WeightedMinkowski | 0.029895s |  0.018471s |  1.6185 |
-| WeightedHamming | 0.013427s |  0.004082s |  3.2896 |
-| SqMahalanobis | 0.121636s |  0.019370s |  6.2796 |
-| Mahalanobis | 0.117871s |  0.019939s |  5.9117 |
+| SqEuclidean | 0.007464s |  0.001993s |  3.7459 |
+| Euclidean | 0.007294s |  0.002035s |  3.5840 |
+| Cityblock | 0.007411s |  0.001991s |  3.7231 |
+| Chebyshev | 0.011455s |  0.005540s |  2.0678 |
+| Minkowski | 0.023713s |  0.016705s |  1.4195 |
+| Hamming | 0.006996s |  0.001985s |  3.5250 |
+| CosineDist | 0.008863s |  0.003301s |  2.6850 |
+| CorrDist | 0.014315s |  0.016693s |  0.8575 |
+| ChiSqDist | 0.008136s |  0.002109s |  3.8578 |
+| KLDivergence | 0.035595s |  0.032610s |  1.0916 |
+| RenyiDivergence(0) | 0.014064s |  0.009562s |  1.4707 |
+| RenyiDivergence(1) | 0.042286s |  0.039520s |  1.0700 |
+| RenyiDivergence(2) | 0.017232s |  0.013540s |  1.2727 |
+| RenyiDivergence(∞) | 0.015273s |  0.010613s |  1.4391 |
+| JSDivergence | 0.085558s |  0.075936s |  1.1267 |
+| BhattacharyyaDist | 0.009933s |  0.004510s |  2.2023 |
+| HellingerDist | 0.010375s |  0.003817s |  2.7184 |
+| WeightedSqEuclidean | 0.007462s |  0.002019s |  3.6957 |
+| WeightedEuclidean | 0.007483s |  0.002075s |  3.6067 |
+| WeightedCityblock | 0.007437s |  0.002034s |  3.6563 |
+| WeightedMinkowski | 0.023364s |  0.017775s |  1.3144 |
+| WeightedHamming | 0.009027s |  0.002754s |  3.2776 |
+| SqMahalanobis | 0.103410s |  0.032976s |  3.1359 |
+| Mahalanobis | 0.104362s |  0.033305s |  3.1336 |
 
-We can see that using ``colwise`` instead of a simple loop yields considerable gain (2x - 6x), especially when the internal computation of each distance is simple. Nonetheless, when the computaton of a single distance is heavy enough (e.g. *Minkowski* and *JSDivergence*), the gain is not as significant.
+We can see that using ``colwise`` instead of a simple loop yields considerable gain (2x - 4x), especially when the internal computation of each distance is simple. Nonetheless, when the computation of a single distance is heavy enough (e.g. *KLDivergence*,  *RenyiDivergence*), the gain is not as significant.
 
 #### Pairwise benchmark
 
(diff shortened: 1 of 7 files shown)


richardreeve:
@ararslan This was intended to fix the issue I addressed in #58 that metricity is not checked and can have implications for how things are calculated... does it make sense to you?

ararslan:
It does indeed, thanks. This looks good to me overall, but you've used 2 space indents in several places--if you could change those to 4 for consistency with the rest of the package, that would be great.

richardreeve:
Ah, thanks for pointing that out. Now fixed.

src/metrics.jl
```diff
@@ -59,14 +59,14 @@ type SpanNormDist <: SemiMetric end
 const UnionMetrics = Union{Euclidean, SqEuclidean, Chebyshev, Cityblock, Minkowski, Hamming, Jaccard, RogersTanimoto, CosineDist, CorrDist, ChiSqDist, KLDivergence, RenyiDivergence, JSDivergence, SpanNormDist}
 
 """
-    Euclidean([thresh])
```
ararslan: This kind of indentation in the docstring is correct and shouldn't be removed

test/test_dists.jl
```diff
@@ -481,17 +481,17 @@ Q = Q * Q'  # make sure Q is positive-definite
...
-    pd = pairwise(Euclidean(1e-12), X)
-    @test pd[1,1] == 0
-    @test pd[2,2] == 0
-    pd = pairwise(SqEuclidean(1e-12), X, X)
-    @test pd[1,1] == 0
-    @test pd[2,2] == 0
-    pd = pairwise(SqEuclidean(1e-12), X)
-    @test pd[1,1] == 0
-    @test pd[2,2] == 0
+X = [0.1 0.2; 0.3 0.4; -0.1 -0.1]
```
ararslan: The contents of `@testset`s should be indented, so this was correct

src/metrics.jl
```diff
@@ -59,14 +59,14 @@ type SpanNormDist <: SemiMetric end
...
+Euclidean([thresh])
 
 Create a euclidean metric.
 
 When computing distances among large numbers of points, it can be much
 more efficient to exploit the formula
 
-    (x-y)^2 = x^2 - 2xy + y^2
+(x-y)^2 = x^2 - 2xy + y^2
```
ararslan: Should should be either reindented or wrapped in double backticks

test/test_dists.jl
```diff
@@ -1,68 +1,153 @@
...
+macro test_metricity(_dist, _x, _y, _z)
+    quote
+        dist = $(esc(_dist))
+        x = $(esc(_x))
+        y = $(esc(_y))
+        z = $(esc(_z))
+        dxy = evaluate(dist, x, y)
+        dxz = evaluate(dist, x, z)
+        dyz = evaluate(dist, y, z)
+        if (isa(dist, PreMetric))
```
ararslan: No parens around conditions

test/test_dists.jl
```diff
@@ -101,28 +182,24 @@ w = rand(size(a))
...
-klv = 0.
+klv = 0.0
 for i = 1 : length(p)
     if p[i] > 0
         klv += p[i] * log(p[i] / q[i])
     end
 end
 @test kl_divergence(p, q) ≈ klv
 
-@test renyi_divergence(p, p, 0) ≈ 0
```
ararslan: Why are these being removed?

richardreeve: Because they are all tested in the Premetric test - Premetrics are non-negative and zero when measuring a distance from something to itself.

test/test_dists.jl
richardreeve: Ah, all of the other `@testset`s were the other way. I've corrected them all now

richardreeve:
Thanks @ararslan - as I say in the comments, the Premetric test includes checks for equality to zero for self-distance, so they don't need to be repeated (as well as many other tests). And all but one of the `@testset`s were not indented originally, but I've now fixed them all (I think!).

src/metrics.jl
```diff
@@ -86,10 +86,10 @@ julia> pairwise(Euclidean(1e-12), x, x)
  0.0
 ```
 """
-Euclidean() = Euclidean(0)
+    Euclidean() = Euclidean(0)
 
 """
-    SqEuclidean([thresh])
+SqEuclidean([thresh])
```
ararslan: Should also be indented

richardreeve:
Oops, thanks.

richardreeve:
Small test fix required as Weighted Hamming turned out to have similar rounding errors to other metrics, resulting in test failure for dxz ≤ dxy + dyz - fixed by allowing dxz ≈ dxy + dyz for Metrics, not just strict equality.

README.md
```diff
@@ -151,7 +151,7 @@ Each distance corresponds to a distance type. The type name and the correspondin
 |  WeightedMinkowski   |  `wminkowski(x, y, w, p)`  | `sum(abs(x - y).^p .* w) ^ (1/p)` |
 |  WeightedHamming     |  `whamming(x, y, w)`       | `sum((x .!= y) .* w)`  |
 
-**Note:** The formulas above are using *Julia*'s functions. These formulas are mainly for conveying the math concepts in a concise way. The actual implementation may use a faster way.
+**Note:** The formulas above are using *Julia*'s functions. These formulas are mainly for conveying the math concepts in a concise way. The actual implementation may use a faster way. x and y arguments are arrays of real numbers; k and l arguments are arrays of distinct elements of any kind; a and b arguments are arrays of `Bool`s; and p and q arguments are arrays forming a discrete probability distribution and are therefore expected to sum to one.
```
KristofferC: Perhaps:

"The arguments `x` and `y` are arrays of real numbers; `k` and `l` are arrays of distinct elements of any kind; `a` and `b` are arrays of `Bool`s; and finally, `p` and `q`  are arrays forming a discrete probability distribution and are therefore both expected to sum to one."

src/metrics.jl
```diff
@@ -86,7 +86,7 @@ julia> pairwise(Euclidean(1e-12), x, x)
  0.0
 ```
 """
-Euclidean() = Euclidean(0)
+    Euclidean() = Euclidean(0)
```
KristofferC: Why indent here?

test/test_dists.jl
```diff
@@ -1,245 +1,332 @@
...
+        z = $(esc(_z))
+        dxy = evaluate(dist, x, y)
+        dxz = evaluate(dist, x, z)
+        dyz = evaluate(dist, y, z)
+        if isa(dist, PreMetric)
+            # Unfortunately small non-zero numbers (~10^-16) are appearing
+            # in our tests due to accumulating floating point rounding errors.
+            # We either need to allow small errors in our tests or change the
+            # way we do accumulations...
+            @test evaluate(dist, x, x) + one(eltype(x)) ≈ one(eltype(x))
```
KristofferC: Doesn't `evaluate(dist, x, x) ≈ 0` work? Otherwise perhaps use `isapprox`?

KristofferC: Also, if you want, this while thing could just be changed to a function instead of a macro.

src/metrics.jl
```diff
@@ -208,46 +208,46 @@ kl_divergence(a::AbstractArray, b::AbstractArray) = evaluate(KLDivergence(), a,
 
 # RenyiDivergence
 function eval_start{T<:AbstractFloat}(::RenyiDivergence, a::AbstractArray{T}, b::AbstractArray{T})
```
KristofferC: Should these be called `p` and `q` like in the README?

src/metrics.jl
```diff
@@ -208,46 +208,46 @@ kl_divergence(a::AbstractArray, b::AbstractArray) = evaluate(KLDivergence(), a,
...
-    zero(T), zero(T)
+    zero(T), zero(T), sum(a), sum(b)
```
KristofferC: If `p` and `q` (`a` and `b`) should sum to 1, why do we need to compute their sum here?

richardreeve: @KristofferC This is because of accumulation errors in `Float64`s - they turn out to be strewn throughout these distance measures unfortunately. sum(a) is often 0.99999999999 or so, and this causes v slight negative distances in some places, and other similar errors. I fixed them here for Renyi, but I don't know think there's a general way of doing it for other metrics, particularly when they are not probability distributions, so they shouldn't sum to 1 anyway.

test/test_dists.jl
richardreeve: No, it doesn't - again there are too many small errors in accumulation of the floating point numbers. The probabilities that these often represent are ≈ 1, so I thought that was a reasonable precision to use - otherwise I'd have been picking some arbitrary error tolerance using approx() that should more reasonably have depended on the floating point precision being used. You're right it could also be put into a function - in fact I had it as that earlier - but I didn't feel it added to the clarity of the code.

src/metrics.jl
ararslan: OH WOW I'M SO DUMB. I thought this was the top of the docstring when I was doing my initial review so I told @richardreeve to indent it. Sorry!

README.md
richardreeve: Fixed.

src/metrics.jl
richardreeve: Oops, getting carried away.

src/metrics.jl
richardreeve: Honestly, these argument names would have to be changed throughout the code for a lot of the distance metrics - I can do that, but it'll look like I wrote the whole package... I wasn't sure it was worth it, but I'm happy to do it if you like?

richardreeve:
@KristofferC I've added in some docs, and addressed your issues (some with revisions, but the impossible-to-avoid problem is rounding errors which caused the others). Is this okay to merge now?

src/metrics.jl
```diff
@@ -29,6 +29,40 @@ type CorrDist <: SemiMetric end
...
+Rényi defined a spectrum of divergence measures generalising the
+Kullback–Leibler divergence (see `KLDivergence`). The divergence is
+not a semimetric as it is not symmetric. It is parameterised by a
+parameter α, and is equal to Kullback–Leibler divergence at α = 1:
+
+At α = 0, ``R_0(P | Q) = -log(sum_{i: p_i > 0}(q_i))``
+
+At α = 1, ``R_1(P | Q) = sum_{i: p_i > 0}(p_i log(p_i / q_i))``
+
+At α = ∞, ``R_∞(P | Q) = log(sup_{i: p_i > 0}(p_i / q_i))``
```
ararslan: In math sections, Unicode literals should be replaced with the equivalent LaTeX, e.g. `\infty` in place of `∞`

richardreeve: Actually, much to my surprise, the [docs](http://docs.julialang.org/en/latest/manual/documentation.html) say the opposite:

```
Use backticks to identify code and equations.

Julia identifiers and code excerpts should always appear between backticks ` to enable
highlighting. Equations in the LaTeX syntax can be inserted between double backticks ``.
Use Unicode characters rather than their LaTeX escape sequence, i.e. ``α = 1`` rather
than ``\\alpha = 1``.
```

I suspect it's because when you do `?RenyiDivergence` at the command line, it can't actually do latex, so you're better off with this bastardised version...

ararslan: Wow, huh. TIL. Okay, I guess this is fine as-is, then.

richardreeve:
@ararslan So, can we go ahead and merge this then?

ararslan:
In light of https://github.com/JuliaStats/Distances.jl/pull/70, could you rebase out the commits that change the benchmarking stuff?
```git
git checkout metricity
git rebase -i HEAD~25  # since there are 25 commits here
```
Then delete commits as needed. That should reduce the amount of branch conflict resolution needed in conjunction with #70.

richardreeve:
@ararslan I have to admit I'm quite hazy on what I did there, but I resolved the conflicts after doing the rebase you suggested. Hopefully that helped.

ararslan:
Okay, that didn't quite work, but that's fine. We can just merge this and rebase #70. Does that sound reasonable to you, @KristofferC?

richardreeve:
@KristofferC - is this good to go then?

PR #72, opened 2017-07-12
Make distances autodiffable

kskyten:
Basically I just changed the `AbstractFloat` types to `Real` for ForwardDiff and ReverseDiff compatibility. The tests still pass. Is there some reason to use the `AbstractFloat` type?

I just did the following quick check for AD functionality.
```julia
using Distances
import ForwardDiff
import ReverseDiff

X = rand(3, 10);
f(w) = sum(pairwise(WeightedSqEuclidean(w), X))
df(w) = ForwardDiff.gradient(f, w)
df2(w) = ReverseDiff.gradient(f, w)

df([1., 2., 3.])
df2([1., 2., 3.])
```
How do I verify that the AD is working properly? ReverseDiff documentation says that mutation of the input is not allowed. There are some updating assignments in the code, but as far as I know these are not mutations in Julia. Is it safe to assume that if I get results, they are correct (i.e. I wont get wrong results silently)?

### DIFF
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -13,7 +13,7 @@ result_type{T}(::SqMahalanobis{T}, ::AbstractArray, ::AbstractArray) = T
 
 # SqMahalanobis
 
-function evaluate{T<:AbstractFloat}(dist::SqMahalanobis{T}, a::AbstractVector, b::AbstractVector)
+function evaluate{T<:Real}(dist::SqMahalanobis{T}, a::AbstractVector, b::AbstractVector)
     if length(a) != length(b)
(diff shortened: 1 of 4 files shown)


KristofferC:
Could put a test dependency on forwarddiff, and use it to test that AD works. Then use either analytic or numeric differentiation to test for correctness.

ararslan:
Bump! Still to do here:

- [ ] Fix conflicts (in particular this will require switching parametric syntaxes)
- [ ] Add correctness tests

Once that's done we'll be good to go. Thanks for the contribution!

KristofferC:
Should be fixed in #74

PR #73, opened 2017-07-25
fix deprecation warnings for new where syntax

KristofferC:
For later

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,16 +1,26 @@
 language: julia
+
 os:
     - linux
     - osx
+
 julia:
-    - 0.5
+    - 0.6
     - nightly
+
+matrix:
+    allow_failures:
+        - julia: nightly
+
 notifications:
     email: false
+
 sudo: false
+
 script:
     - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
     - julia -e 'Pkg.clone(pwd()); Pkg.build("Distances")'
     - julia -e 'Pkg.test("Distances", coverage=true)'
+
 after_success:
     - julia -e 'cd(Pkg.dir("Distances")); Pkg.add("Coverage"); using Coverage; Coveralls.submit(process_folder())'
--- a/REQUIRE
+++ b/REQUIRE
@@ -1,2 +1,1 @@
-julia 0.5
-Compat 0.18
+julia 0.6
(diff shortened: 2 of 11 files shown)


KristofferC:
You ok with this @ararslan?

.travis.yml
```diff
@@ -1,16 +1,26 @@
...
     - linux
     - osx
+
 julia:
-    - 0.5
+    - 0.6
     - nightly
+
+matrix:
+    allow_failures:
```
ararslan: Why are we allowing 0.7 failures?

ararslan:
Yep, still looks good to me aside from the `allow_failures` question.

.travis.yml
KristofferC: Because it isn't released and badges / PRs etc shouldn't be red because Base Julia decided to deprecate `+` or something.

ararslan: I'd tend to disagree; I think failures due to changes in Base should be caught and addressed as they arise, and CI failures are pretty much the only reliable way to get that. But I don't spent much time maintaining this package, so ¯\\\_(ツ)\_/¯

KristofferC: Showing a big fat red "BUILD FAILING"-badge as the first thing a user sees because Base has changed is lame :P. It is, of course, good to have the package work on master but failing the build is excessive in my opinion.

ararslan: Okay, I'll leave it up to you

richardreeve:
@KristofferC It took me a bit of time to spot this since there were no test errors, but converting my test_metricity macro to a function has removed all of those (hundreds of tests) tests. This is [the actual test log](https://travis-ci.org/JuliaStats/Distances.jl/jobs/259404124) from this branch - you can see that there are now no metricity tests. Master has the same [problem](https://travis-ci.org/JuliaStats/Distances.jl/jobs/308682362). I'll put in a PR to fix this.

KristofferC:
Well, the tests are still running, they just don't get added to the testset.

richardreeve:
No they aren't - now that I've added them in several fail. In fact I've now identified what has happened both here and in the pairwise tests. The tests were `quote`d from being a macro, so were never evaluated. This is now fixed in the PR, but I am now fighting errors in the newly run tests.

PR #74, opened 2017-08-01
relax metrics to Real

KristofferC:
Also tests it by defining a new number type that is a subtype of Real and calling everything.

Replaces #72.

### DIFF
--- a/src/Distances.jl
+++ b/src/Distances.jl
@@ -58,7 +58,6 @@ export
     rogerstanimoto,
     chebyshev,
     minkowski,
-    mahalanobis,
 
     hamming,
     cosine_dist,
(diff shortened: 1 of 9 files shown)


src/common.jl
```diff
@@ -65,7 +65,7 @@ end
 
 function get_colwise_dims(d::Int, r::AbstractArray, a::AbstractMatrix, b::AbstractVector)
     size(a, 1) == length(b) == d
-        throw(DimensionMismatch("Incorrect vector dimensions."))
```
KristofferC: Thank you formatter, this was a missed `||` here.

src/Distances.jl
```diff
@@ -58,7 +58,6 @@ export
     rogerstanimoto,
     chebyshev,
     minkowski,
-    mahalanobis,
```
KristofferC: already exported above

Issue #75, opened 2017-08-24
Implement Bray-Curtis dissimilarity?


kescobo:
[Bray-Curtis](https://en.wikipedia.org/wiki/Bray%E2%80%93Curtis_dissimilarity) is a semi-metric often used in ecology.

The implementation for two vectors i and j is pretty straightforward:

```julia
function braycurtis(i::AbstractVector, j::AbstractVector)
    length(i) == length(j) || error("Vectors must be same length")
    return sum(abs.(i .- j))/sum(i .+ j)
end
```

I was looking around in the source code, and I'd be happy to try to submit a PR but I'm not entirely confident I understand what's going on with all of the` @inbounds` and `@eval` stuff...

kescobo:
Actually, I [gave it a shot](https://github.com/kescobo/Distances.jl/tree/braycurtis) by copying the form of the Jaccard code, and it works! I'm not sure what the `@eval_reduce` block does, so I left that alone, but I tested it against some of my data and it seems to work as expected for two vectors and for `pairwise()`. 

Please let me know if you'd be interested in a PR.

ararslan:
Yeah, a PR would be great!

PR #76, opened 2017-08-24
Add BrayCurtis

kescobo:
Addresses #75 

I modeled this off of the jaccard code, since they're relatively similar. 

This still needs tests, but I could use some guidance as to whether the way I've done this makes sense, and what tests I should add.

### DIFF
--- a/README.md
+++ b/README.md
@@ -140,7 +140,7 @@ Each distance corresponds to a distance type. The type name and the correspondin
 |  Hamming             |  `hamming(k, l)`           | `sum(k .!= l)` |
 |  RogersTanimoto      |  `rogerstanimoto(a, b)`    | `2(sum(a&!b) + sum(!a&b)) / (2(sum(a&!b) + sum(!a&b)) + sum(a&b) + sum(!a&!b))` |
 |  Jaccard             |  `jaccard(x, y)`           | `1 - sum(min(x, y)) / sum(max(x, y))` |
-|  BrayCurtis          |  `braycurtis(x, y)`        | `sum(abs(x - y)) / sum(abs(x+y))`  |
+|  BrayCurtis          |  `braycurtis(x, y)`        | `sum(abs(x - y)) / sum(abs(x + y))`  |
 |  CosineDist          |  `cosine_dist(x, y)`       | `1 - dot(x, y) / (norm(x) * norm(y))` |
 |  CorrDist            |  `corr_dist(x, y)`         | `cosine_dist(x - mean(x), y - mean(y))` |
 |  ChiSqDist           |  `chisq_dist(x, y)`        | `sum((x - y).^2 / (x + y))` |
--- a/README.md
+++ b/README.md
@@ -140,10 +140,7 @@ Each distance corresponds to a distance type. The type name and the correspondin
 |  Hamming             |  `hamming(k, l)`           | `sum(k .!= l)` |
 |  RogersTanimoto      |  `rogerstanimoto(a, b)`    | `2(sum(a&!b) + sum(!a&b)) / (2(sum(a&!b) + sum(!a&b)) + sum(a&b) + sum(!a&!b))` |
 |  Jaccard             |  `jaccard(x, y)`           | `1 - sum(min(x, y)) / sum(max(x, y))` |
-<<<<<<< master
 |  BrayCurtis          |  `braycurtis(x, y)`        | `sum(abs(x - y)) / sum(abs(x+y))`  |
-=======
->>>>>>> @alyst review fixes
 |  CosineDist          |  `cosine_dist(x, y)`       | `1 - dot(x, y) / (norm(x) * norm(y))` |
 |  CorrDist            |  `corr_dist(x, y)`         | `cosine_dist(x - mean(x), y - mean(y))` |
 |  ChiSqDist           |  `chisq_dist(x, y)`        | `sum((x - y).^2 / (x + y))` |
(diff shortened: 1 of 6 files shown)


KristofferC:
TODO: Test correctness on some points including some degenerate cases,  and adding the distance to the table in the README. Should be good after that,

kescobo:
@KristofferC Ahh, sorry I missed the table - fixed. I presume someone else will run the benchmarks so it's consistant with the others? 

Would you please clarify what you mean by "degenerate cases?" I added a couple of things that I think are along those lines... does that satisfy what you meant, or was there something else?

nalimilan:
@KristofferC Bump.

kescobo:
Thanks for the bump @nalimilan :-)

@KristofferC I'm hoping to do an initial release of a microbiome package that's somewhat dependent on this PR. Any chance you could review the test cases I added are sufficient?

README.md
```diff
@@ -128,37 +129,38 @@ This type system has practical significance. For example, when computing pairwis
...
+|  SpanNormDist             |  `spannorm_dist(x, y)`     | `max(x - y) - min(x - y )` |
+|  BhattacharyyaDist        |  `bhattacharyya(x, y)`     | `-log(sum(sqrt(x .* y) / sqrt(sum(x) * sum(y)))` |
+|  HellingerDist            |  `hellinger(x, y) `        | `sqrt(1 - sum(sqrt(x .* y) / sqrt(sum(x) * sum(y))))` |
+|  Mahalanobis              |  `mahalanobis(x, y, Q)`    | `sqrt((x - y)' * Q * (x - y))` |
+|  SqMahalanobis            |  `sqmahalanobis(x, y, Q)`  | ` (x - y)' * Q * (x - y)`  |
+|  MeanAbsDeviation         |  `meanad(x, y)`            | `mean(abs.(x - y))` |
+|  MeanSqDeviation          |  `msd(x, y)`               | `mean(abs2.(x - y))` |
+|  RMSDeviation             |  `rmsd(x, y)`              | `sqrt(msd(x, y))` |
+|  NormRMSDeviation         |  `nrmsd(x, y)`             | `rmsd(x, y) / (maximum(x) - minimum(x))` |
+| Bray-Curtis Dissimliarity |  `braycurtis(x, y)`        | `sum(abs(x - y)) / sum(abs(x+y))`  |
```
alyst: IIUC, it should be just BrayCutris, since in this column you provide the type names to be used from e.g. `pairwise()` method.
So increasing the column width would be also unnecessary.

kescobo: @alyst Sorry for delayed response (holiday). I'll take a look at this today.

test/test_dists.jl
```diff
@@ -169,6 +169,13 @@ end
         @inferred evaluate(Jaccard(), [1, 2, 3], [1, 2, 3])
         @inferred evaluate(Jaccard(), [true, false, true], [false, true, true])
 
+        # Test Bray-Curtis. Should be 1 if no elements are shared, 0 if all
+        # are the same
+        @test braycurtis([1,0,3],[0,1,0]) == 1.0
+        @test braycurtis(rand(10), zeros(10)) == 1.0
+        @test braycurtis([1,0],[1,0]) == 0.0
+
```
alyst: I guess `Bool` case has to be tested as well.
Also, there's `T`, `_x`, `_y` loop above, I think it makes sense to add `braycurtis()` tests there as well (that should test `Float64` case).

src/metrics.jl
```diff
@@ -356,6 +357,27 @@ end
...
+    abs_p = abs(s1 + s2)
+    abs_m, abs_p
+end
+@inline function eval_reduce(::BrayCurtis, s1, s2)
+    @inbounds a = s1[1] + s2[1]
+    @inbounds b = s1[2] + s2[2]
+    a, b
+end
+@inline function eval_end(::BrayCurtis, a)
+    @inbounds v = a[1] / a[2]
```
alyst: Should `a[1] == 0 && a[2] == 0` case give NaN or 0.0?

README.md
kescobo: Fixed.

test/test_dists.jl
kescobo: I added `braycurtis()` to the loop, as suggested, but I don't think Bray Curtis is defined for the boolean case, or at least it's not really an appropriate dissimilarity measure for presence/absence type situations.

src/metrics.jl
kescobo: It should give NaN (or throw an error) - rows where both counts are zero should not be included in the calculation.

kescobo:
Thanks for the review @alyst - I've made the fixes you suggested, except for adding a test for the boolean case, as I don't think that would be an appropriate use of this distance metric. If you think it should be supported anyway, I'd be happy to add it in.

test/test_dists.jl
alyst: You have defined `eval_start()` for a `Bool` case, so I supposed it have to be tested.
E.g `Euclidean` or `Cityblock` also support boolean vectors.

kescobo:
@alyst OK - I added a test for the boolean case and rebased. Tests passed locally, so I'm guessing travis will give the green light. Thanks again for taking this on!

alyst:
Yes, I guess it would be better to update the benchmarks in README.md using the same machine.
But meanwhile you can add BrayCurtis to the benchmark scripts and post the results here.

kescobo:
I've updated the benchmark files, but was unable to complete the run. Complete output [is here](https://gist.github.com/kescobo/c2de44e7bee52382c05f94c86c6bfe30), but the relevant error is:

```
ERROR: LoadError: ArgumentError: JLD serialization is no longer supported. Benchmarks should now be saved in
JSON format using `save("params".json, args...)` and loaded from JSON using
using `load("params".json, args...)`. You will need to convert existing
saved benchmarks to JSON in order to use them with this version of BenchmarkTools.
Stacktrace:
 [1] badext(::String) at /Users/kev/.julia/v0.6/BenchmarkTools/src/serialization.jl:61
 [2] save(::String, ::String, ::Vararg{Any,N} where N) at /Users/kev/.julia/v0.6/BenchmarkTools/src/serialization.jl:65
 [3] include_from_node1(::String) at ./loading.jl:576
 [4] include(::String) at ./sysimg.jl:14
 [5] process_options(::Base.JLOptions) at ./client.jl:305
 [6] _start() at ./client.jl:371
while loading /Users/kev/.julia/v0.6/Distances/benchmark/print_table.jl, in expression starting on line 43
```

kescobo:
@alyst Seems like the above error is a problem with the benchmarking script rather than with the PR, right? I think this is ready to go.

alyst:
@kescobo Yes, I'm just waiting to collect feedback :) Otherwise I'll merge it around 22:00 CET

PR #77, opened 2017-09-29
Fix #71: Add Haversine distance

### LINKED ISSUE #71: Implement Haversine
Haversine is a very useful distance when dealing with geospatial data. It is implemented in libraries such as scikit-learn, and it's pretty stardard today.

juliohm:
Fix #71

The Haversine formula for 2D points on a sphere: https://en.wikipedia.org/wiki/Haversine_formula

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -10,13 +10,20 @@ struct Haversine{T<:Real} <: Metric
     radius::T
 end
 
-function evaluate(dist::Haversine{T}, x::AbstractVector, y::AbstractVector) where {T<:Real}
-    # longitudes
-    Δλ = deg2rad(y[1] - x[1])
+const VecOrLengthTwoTuple{T} = Union{AbstractVector{T}, NTuple{2, T}}
 
-    # latitudes
-    φ₁ = deg2rad(x[2])
-    φ₂ = deg2rad(y[2])
+function evaluate(dist::Haversine, x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple) 
+    length(x) == length(y) == 2 || haversine_error()
+
+    @inbounds begin
+        # longitudes
+        Δλ = deg2rad(y[1] - x[1])
+
+        # latitudes
+        φ₁ = deg2rad(x[2])
+        φ₂ = deg2rad(y[2])
+    end
+
     Δφ = φ₂ - φ₁
 
     # haversine formula
@@ -27,4 +34,6 @@ function evaluate(dist::Haversine{T}, x::AbstractVector, y::AbstractVector) wher
     c*dist.radius
 end
 
-haversine(x::AbstractVector, y::AbstractVector, radius::T) where {T<:Real} = evaluate(Haversine(radius), x, y)
+haversine(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple, radius::Real) = evaluate(Haversine(radius), x, y)
+
+@noinline haversine_error() = throw(ArgumentError("expected both inputs to have length 2 in Haversine distance"))
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -39,3 +39,4 @@ function evaluate(dist::Haversine{T}, x::AbstractVector, y::AbstractVector) wher
 end
 
 haversine(x::AbstractVector, y::AbstractVector, radius::T) where {T<:Real} = evaluate(Haversine(radius), x, y)
+haversine(x::AbstractVector, y::AbstractVector) = evaluate(Haversine(), x, y)
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -15,27 +15,27 @@ the Earth as a sphere is typically negligible for most applications. It is
 no more than 0.3%.
 """
 struct Haversine{T<:Real} <: Metric
-  radius::T
+    radius::T
 end
 
 # use Earth radius ≈ 6371km by default
 Haversine() = Haversine(6371.)
 
 function evaluate(dist::Haversine{T}, x::AbstractVector, y::AbstractVector) where {T<:Real}
-  # longitudes
-  Δλ = deg2rad(y[1] - x[1])
+    # longitudes
+    Δλ = deg2rad(y[1] - x[1])
 
-  # latitudes
-  φ₁ = deg2rad(x[2])
-  φ₂ = deg2rad(y[2])
-  Δφ = φ₂ - φ₁
+    # latitudes
+    φ₁ = deg2rad(x[2])
+    φ₂ = deg2rad(y[2])
+    Δφ = φ₂ - φ₁
 
-  # haversine formula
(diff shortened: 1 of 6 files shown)


juliohm:
This is ready for review.

juliohm:
Could someone please review this PR? I have other distances that I am planning to add after this is merged. The ultimate goal is to make GeoStats.jl depend on Distances.jl.

src/haversine.jl
```diff
@@ -0,0 +1,42 @@
...
+Locations are described with longitude and latitude in degrees and
+the radius of the Earth is used by default (≈ 6371km). The computed
+distance has the same units as that of the radius.
+
+### Notes
+
+The haversine formula is widely used to approximate the geodesic distance
+between two points at the surface of the Earth. The error from approximating
+the Earth as a sphere is typically negligible for most applications. It is
+no more than 0.3%.
```
ararslan: This is an interesting anecdote but it seems too specific to have in a package as general as Distances.

src/haversine.jl
```diff
@@ -0,0 +1,42 @@
...
+"""
+struct Haversine{T<:Real} <: Metric
+  radius::T
```
ararslan: Please use four spaces for indentation in keeping with the rest of the repo. In general, the stylistic conventions in JuliaStats tries to mirror those in Base.

src/haversine.jl
```diff
@@ -0,0 +1,42 @@
...
+end
+
+# use Earth radius ≈ 6371km by default
+Haversine() = Haversine(6371.)
```
ararslan: I really don't think we should define this here. As with the note in the docstring, it's too specific for such a general package, and in this case I think it's highly error-prone.

src/haversine.jl
juliohm: I can remove the comment @ararslan if you like, should I?

ararslan: I think it would be good to remove it

src/haversine.jl
juliohm: @ararslan I think by far the application of the Haversine is related to measurements on our planet, few people will use it with a different radius. If it is necessary to remove the default radius, I can do it, I just don't see this as a restriction, it is a default value only.

juliohm: Elaborating on this, when people talk about this distance they are usually thinking about geosciences. I personally don't think we are missing any generality here as the user can still use any radius he/she wants.

test/test_dists.jl
```diff
@@ -272,6 +278,13 @@ end # testset
     end
 end #testset
 
+@testset "haversine" begin
+  for T in (Float64, F64)
```
ararslan: This block needs adjustment as well

juliohm: Thanks, missed that...

juliohm:
@ararslan all the requested changes should be implemented now, let's see if the tests pass. Please let me know if I missed something else.

ararslan:
Thanks for your great work here and for being so responsive to feedback. 👍

juliohm:
Tests passed :+1:

src/haversine.jl
```diff
@@ -0,0 +1,30 @@
...
+The haversine distance between two locations on a sphere of given `radius`.
+
+Locations are described with longitude and latitude in degrees.
+The computed distance has the same units as that of the radius.
+"""
+struct Haversine{T<:Real} <: Metric
+    radius::T
+end
+
+function evaluate(dist::Haversine{T}, x::AbstractVector, y::AbstractVector) where {T<:Real}
```
KristofferC: Perhaps check that the length is 2? Also, I think it would be nice to accept tuples of length 2 as well.

juliohm: Can this check be performed without performance penalties? That is the main reason I didn't write a test. Also, could you please clarify the interface with tuples you have in mind?

KristofferC: My thinking was that if you have a bunch of longitudes and latitudes in a list you would have to preallocate an array to send into this function in order to not cause an allocation at every call. You could use a static array but then you need a package. With a tuple, you won't have any allocations.

Regarding performance time, a check like this should be completely negligible. But you could of course benchmark it.

juliohm: @KristofferC is this issue about pre-allocation specific to Haversine or to the Distances.jl API in general? I didn't write a method with tuples because I didn't see any with the other metrics defined in the package.

Regarding the check for 2D vectors, I prefer to not implement it because I am expecting to call this function some hundreds of thousands of times, I had a performance boost on my package already by just eliminating basic assertions. A user of this distance will know that it is only supposed to make sense with 2D vectors.

juliohm:
@ararslan what is the effect of adding Haversine to `UnionMetrics`? I thought this Union type existed only for implementations of metrics that can be written in terms of the same components like `eval_op`, `eval_end`, etc. Haversine doesn't fit into this recipe.

src/haversine.jl
KristofferC: If you call this function hundreds of thousands of times, the check will have cost you ~0.0001 seconds. However, if you do the length check you can add `@inbounds` on the array access, giving you back that precious nanosecond every call.

Furthermore, if you allow this function to take tuples, you will gain 2 more nanoseconds per call compared to the arrays.

To summarize:

* Doing the check makes the function faster at no performance cost.
* Since you are so worried about performance, allowing tuples will allow you to call this function with better performance.

ararslan:
Ah, okay. Then disregard `UnionMetrics` then.

Also it seems this that this function assumes the input is in degrees?

src/haversine.jl
juliohm: @KristofferC I appreciate your suggestions, but the use case with tuples is not very appealing to me at the moment. All my use cases involve arrays coming from other calculations that aren't necessarily 2D. So specializing on 2D tuples is not a common pattern for me. Also, by grepping the source code of Distances.jl, the only other type using tuples is the RenyDivergence. All other metrics have their implementations with arrays.

Thank you again for your suggestions, but I think a version of Haversine with tuples could be added in the future if someone misses it.

juliohm:
@ararslan yes, the function expects latitude and longitude values in degrees. Any other change that you would like to see before the merge?

ararslan:
It seems that radians would be a more appropriate default than degrees. Or perhaps an option to select one or the other?

src/haversine.jl
juliohm: Please do, I am not sure I understand what your suggestions are here, but if you can update the PR to be more efficient and satisfy a general interface (that also works with arrays), that would be great.

juliohm:
Thank you @KristofferC, the union type you wrote is nice. I'd say the convention is to pass arguments in degrees, but maybe one could have an option as @ararslan suggested for switching between degrees and radians?

Would it make sense to have a field in Haversine that stores the type of the arguments? `:rad` or `:deg`? And introduce a branch in the body of the function evaluate? I wonder how this option could be incorporated without introducing branches since we don't have a type for `degree` vs. `radian`.

PR #80, opened 2017-10-27
Handle floating point errors in Haversine distance

juliohm:
This PR takes care of well-known floating point errors in the haversine formula. It was originally captured here: https://github.com/juliohm/GeoStats.jl/issues/14

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -12,7 +12,7 @@ end
 
 const VecOrLengthTwoTuple{T} = Union{AbstractVector{T}, NTuple{2, T}}
 
-function evaluate(dist::Haversine, x::VecOrLengthTwoTuple{T}, y::VecOrLengthTwoTuple{T}) where {T}
+function evaluate(dist::Haversine, x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple)
     length(x) == length(y) == 2 || haversine_error()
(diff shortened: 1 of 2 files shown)


src/haversine.jl
```diff
@@ -12,7 +12,7 @@ end
 
 const VecOrLengthTwoTuple{T} = Union{AbstractVector{T}, NTuple{2, T}}
 
-function evaluate(dist::Haversine, x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple) 
+function evaluate(dist::Haversine, x::VecOrLengthTwoTuple{T}, y::VecOrLengthTwoTuple{T}) where {T}
```
nalimilan: Do you really need to restrict both vectors to be of the same type? You could just call `one(a)` below, which would be more correct anyway.

juliohm: Very good point, I will update the PR accordingly.

nalimilan: Thanks. Somebody else will have to review the math as I have no idea what this is doing...

juliohm: Sometimes due to floating point errors this variable "a" becomes something like "a = 1.000000001". This change is only discarding the noise.

juliohm:
Could someone please merge this fix?

PR #81, opened 2017-10-28
RFC: Add Ellipsoidal distance

### LINKED ISSUE #78: Ellipsoid distance
The Mahalanobis distance is very useful when we have a covariance/precision matrix to weight the observations. In many other applications where a covariance doesn't exist, however; we still have the opportunity to introduce a geometrical view by constructing a matrix based on rotation and scaling of the system axes.

The Ellipsoid distance I implemented  in GeoStats.jl does exactly this: https://github.com/juliohm/GeoStats.jl/blob/master/src/distances.jl#L38-L105

It is basically a helper constructor that given directions and scaling coefficients, builds the corresponding Mahalanobis matrix. Can I submit a PR that adds this distance to the package? My plan is to just reuse the code for Mahalanobis. The new struct Ellipsoid would just contain a Mahalanobis that is initialized properly.

Please let me know if that is welcome, and I will work on it during the week.

juliohm:
Fix #78.

- [x] Code
- [x] Tests
- [x] Docs
- [x] Benchmarks


juliohm:
What is the purpose of the type F64 in the tests? The inner constructor of Ellipsoidal distance calls cos(F64()) which returns a StackOverflow.

juliohm:
Could you please comment on this initial draft? Should we have a new type called Ellipsoidal or just add more constructors to Mahalanobis?

juliohm:
I am thinking about renaming it to EllipsoidDist instead, I appreciate if someone can review it. This is the last distance I am porting from GeoStats.jl before making Distances.jl a dependency.

KristofferC:
The purpose of F64 is to test the distance functions with a non standard type to make sure no assumptions about the input type being e.g. `Float64` is made.

juliohm:
Thank you @KristofferC , I have added `cos` and `sin` to F64 in order to include more tests for the proposed distance. Could you please review the rest of the PR?

Also, I am finding it difficult to include the distance in the benchmark suite. For this geometrical view I only implement 2D and 3D spaces. The benchmark calls distances with N-dimensional vectors. Any suggestions on how to adapt the script?

juliohm:
I really appreciate if someone can take a look at this, I depend on it to release a new version of my package. Benchmarks are missing and I am open for suggestions. Also, let me know if the name of the distance is clear enough.

KristofferC:
>  I depend on it to release a new version of my package

That seems quite unnecessary. You can just define this distance in your package and release it today, instead.

juliohm:
True, I was just wondering if I could avoid this hack, it doesn't seem very smart since everything is ready here.

KristofferC:
It is definitely not a hack but a feature of Julia that you can define your own types and methods on those types even when the type subtypes something the package does not own.

juliohm:
Any feedback on how to adapt the benchmarks for distances that are only expecting at most 3 dimensions?

juliohm:
- Fixed the benchmarks to include cases for distances defined on 2D and 3D spaces.

- Fxed the persistence format from *.jld to *.json as suggested by the deprecation warning from BenchmarkTools.jl

PR #82, opened 2017-11-22
Fix ChiSq distance around zero

alyst:
Currently `chisq_dist([0.], [0.])` would generate `NaN`, but we can [continuously] define it (0, 0) to be 0.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -239,7 +239,7 @@ corr_dist(a::AbstractArray, b::AbstractArray) = evaluate(CorrDist(), a, b)
 result_type(::CorrDist, a::AbstractArray, b::AbstractArray) = result_type(CosineDist(), a, b)
 
 # ChiSqDist
-@inline eval_op(::ChiSqDist, ai, bi) = abs2(ai - bi) / (ai + bi)
+@inline eval_op(::ChiSqDist, ai, bi) = ifelse(ai != bi, abs2(ai - bi) / (ai + bi), zero(ai))
 @inline eval_reduce(::ChiSqDist, s1, s2) = s1 + s2
 chisq_dist(a::AbstractArray, b::AbstractArray) = evaluate(ChiSqDist(), a, b)
 
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -239,7 +239,7 @@ corr_dist(a::AbstractArray, b::AbstractArray) = evaluate(CorrDist(), a, b)
 result_type(::CorrDist, a::AbstractArray, b::AbstractArray) = result_type(CosineDist(), a, b)
 
 # ChiSqDist
-@inline eval_op(::ChiSqDist, ai, bi) = ifelse(ai != bi, abs2(ai - bi) / (ai + bi), zero(ai))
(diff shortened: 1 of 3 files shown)


src/metrics.jl
```diff
@@ -239,7 +239,7 @@ corr_dist(a::AbstractArray, b::AbstractArray) = evaluate(CorrDist(), a, b)
 result_type(::CorrDist, a::AbstractArray, b::AbstractArray) = result_type(CosineDist(), a, b)
 
 # ChiSqDist
-@inline eval_op(::ChiSqDist, ai, bi) = abs2(ai - bi) / (ai + bi)
+@inline eval_op(::ChiSqDist, ai, bi) = ifelse(ai != bi, abs2(ai - bi) / (ai + bi), zero(ai))
```
nalimilan: Shouldn't the check be `ifelse(ai + bi != 0, ...)`? Also, it's not easy to ensure type stability (e.g. for integer inputs), maybe use something like `zero(ai - bi)/one(ai - bi)`?

alyst: > Shouldn't the check be `ifelse(ai + bi != 0, ...)`?

We can define it as 0 only at `(0, 0)`.
For the neighbourhood of `(x, -x)` line (points like (x, -x + z), where z -> 0) the function is `(2x+z)^2 / z`, i.e. either -> -Inf or -> +Inf depending on `z` sign. So I think it should stay NaN everywhere except (0,0) where it behaves as `z`.

> maybe use something like `zero(ai - bi)/one(ai - bi)`?

+1

src/metrics.jl
```diff
@@ -239,7 +239,7 @@ corr_dist(a::AbstractArray, b::AbstractArray) = evaluate(CorrDist(), a, b)
 result_type(::CorrDist, a::AbstractArray, b::AbstractArray) = result_type(CosineDist(), a, b)
 
 # ChiSqDist
-@inline eval_op(::ChiSqDist, ai, bi) = abs2(ai - bi) / (ai + bi)
+@inline eval_op(::ChiSqDist, ai, bi) = ifelse(ai != bi, abs2(ai - bi) / (ai + bi), zero(abs2(ai - bi) / (ai + bi)))
```
nalimilan: If you use the same expression twice, better store the result in a variable and call `zero` on it when `ai != bi`.

alyst: The 2nd occurrence is only for type inference, it's not calculated (checked on v0.6.1).
I've tried your suggestion: `@code_native` shows that in this particular case the compiler was smart to reorder the operations and eliminate the temporary var, so nothing is put to the stack.
But I would rather not tempt the compiler in performance critical methods like this.

nalimilan: My suggestion was about readability. AFAICT it could only be faster than repeating the expression, and never slower.

alyst: I've implemented your suggestion, because in this case the compiler still generates the optimal code.
I'm just not sure we both agreed that the 2nd occurrence of the formula is eliminated during type inference, so the current version (`d=...; ifelse(ai != bi, d, 0)`) have suboptimal evaluation order, and we rely on LLVM to fix it.

Issue #83, opened 2017-11-22
colwise ridiculously slow for small columns due to allocating array views


chethega:
A quite common need is to have one (or two) data matrices, and wanting to compute distances between certain columns. Using Distances.jl is nice, because it supplies an API, so that users can plug different distance functions into an algorithm. As example, consider NearestNeighbors.jl.

Hence, one needs an API to compute evaluate(dist, X[:,i1], Y[:,i2]).

Unfortunately, Distances.jl uses array views. Array views are harmful, because they allocate (and add indirection!). Until this is fixed, one must provide a zero-overhead way of computing such distances.

A simple API would be evaluate(dist, X, i1, Y, i2). Due to limitations of julia, we cannot bundle the underlying matrix and the indices into a struct or tuple (this is what array views do). This constraint is unfortunate, but I don't think there is any way to have a nice and fast API. As far as I understood, this will not change in julia 1.0; hence an ugly-but-fast API is necessary (if you consider allocating array views a bug, then a workaround is needed). 

I am not sure about the final API design; hence no pull request. However, I expect that a lot of packages downstream will want to make use of this.

Indeed, the broken API is used even internally in colwise. 

I attached an example benchmark, comparing colwise against a naive loop for euclidean distances between 10-dimensional points. Lower dimensional data has even worse relative speed differences; higher dimensional data reduces the difference. 

```
using Distances
using BenchmarkTools

m = 10
n=100_000
Xcol = rand(m,n)
Ycol=rand(m,n)


function colwise_naive(X,Y)
    m = size(X,1)
    n = size(X,2)
    T = eltype(X)
    assert(m == size(Y,1) && n == size(Y,2) && T==eltype(Y))

    res = Vector{T}(n)
    @inbounds  for j = 1:n
        s_ = zero(T)
        @inbounds @simd for i = 1:m
            dd_ = X[i,j]-Y[i,j]
            s_ += dd_*dd_
        end
        res[j] = sqrt(s_)
    end
    res
end

dd = Euclidean()
bench_colw = @benchmark colwise(dd, Xcol,Ycol)
bench_colw2 = @benchmark colwise_naive(Xcol,Ycol)

println("distances.jl colwise")
show(STDOUT, MIME"text/plain"(), bench_colw)

println("\nnaive colwise")
show(STDOUT, MIME"text/plain"(), bench_colw2)

```
yielding
```
colwise
BenchmarkTools.Trial: 
  memory estimate:  9.92 MiB
[...]

PR #86, opened 2017-11-28
use unsafe_views and remove branch for arrays

KristofferC:
An alternative for solving https://github.com/JuliaStats/Distances.jl/issues/83

```
julia> println("distances.jl colwise")
distances.jl colwise

julia> show(STDOUT, MIME"text/plain"(), bench_colw)
BenchmarkTools.Trial:
  memory estimate:  781.33 KiB
  allocs estimate:  2
  --------------
  minimum time:     1.082 ms (0.00% GC)
  median time:      1.256 ms (0.00% GC)
  mean time:        1.364 ms (2.21% GC)
  maximum time:     3.484 ms (24.77% GC)
  --------------
  samples:          3644
  evals/sample:     1
julia> println("\nnaive colwise")

naive colwise

julia> show(STDOUT, MIME"text/plain"(), bench_colw2)
BenchmarkTools.Trial:
  memory estimate:  781.33 KiB
  allocs estimate:  2
  --------------
  minimum time:     1.072 ms (0.00% GC)
  median time:      1.275 ms (0.00% GC)
  mean time:        1.354 ms (2.25% GC)
  maximum time:     3.658 ms (32.67% GC)
  --------------
  samples:          3670
  evals/sample:     1
```

cc @chethega

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -144,7 +144,24 @@ SqEuclidean() = SqEuclidean(0)
 #
 ###########################################################
 
-function evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray)
+# Specialized for Arrays and avoids a branch on the size
+@inline function evaluate(d::UnionMetrics, a::T, b::T) where T <: Union{Array, UnsafeVectorView}
+    if length(a) != length(b)
+        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
+    end
+    if length(a) == 0
+        return zero(result_type(d, a, b))
+    end
+    s = eval_start(d, a, b)
+    @simd for I in eachindex(a, b)
+        @inbounds ai = a[I]
+        @inbounds bi = b[I]
+        s = eval_reduce(d, s, eval_op(d, ai, bi))
+    end
+    return eval_end(d, s)
+end
+
+@inline function evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray)
     if length(a) != length(b)
         throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
     end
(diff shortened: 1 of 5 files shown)


src/metrics.jl
```diff
@@ -144,7 +144,24 @@ SqEuclidean() = SqEuclidean(0)
 #
 ###########################################################
 
-function evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray)
+# Specialized for Arrays and avoids a branch on the size
+@inline function evaluate(d::UnionMetrics, a::T, b::T) where T <: Union{Array, UnsafeVectorView}
```
KristofferC: Note this `@inline`. It apparently does quite a lot for small arrays and it isn't that much code to inline.

chethega:
I think the unsafe_views beats the split-view API in terms of code-readablity and should be equivalent in terms of speed. 

I will try to port over the better memory access / multithreaded pairwise!, as well as the modified boundscheck-propagation from my PR; then my the split-views can be dropped.

Question: I think distances.jl is unhappy with one-dimensional indices not running from 1:length(V)? As in e.g. offset-arrays?

KristofferC:
Yeah, this package was created a long time ago  and the core has been largely unchanged. It is likely it doesn't work well with non one based arrays. We should fix that and test it.

chethega:
You can find the more memory-friendly iteration order on [https://gist.github.com/chethega/00241c6312d3fe28afc5532708a69b3d](https://gist.github.com/chethega/00241c6312d3fe28afc5532708a69b3d). 

After some testing, I could not see any significant performance gains; therefore, it probably does not belong in Distances.jl.

KristofferC:
On https://github.com/JuliaLang/julia/pull/24113, with normal `view`s, I see no allocations as long as I inline `evaluate`. Performance is the same as the hand rolled one as well.

chethega:
Sure. Then, instead of exporting and cementing the fastview, export and document now something like

`@inline @Base.propagate_inbounds evaluate(d, A, jA, B, jB)= evaluate(d, fastview(A,:,jA), fastview(A,:,jB))`

and, after removal of fastview, replace it by 

`@inline @Base.propagate_inbounds evaluate(d, A, jA, B, jB)= evaluate(d, view(A,:,jA),view(A,:,jB))`,

and possibly deprecate this a couple versions later? Then one should maybe add boundscheck to fastview, and be happy.

PR #87, opened 2017-11-28
Unsafe views

chethega:
minor changes in boundscheck-management.

### DIFF
--- a/src/Distances.jl
+++ b/src/Distances.jl
@@ -16,6 +16,8 @@ export
     pairwise!,
     evaluate,
 
+    fastview,
+
     # distance classes
     Euclidean,
     SqEuclidean,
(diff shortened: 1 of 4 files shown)


chethega:
...sorry, I failed at the github UI, this belongs into [https://github.com/JuliaStats/Distances.jl/pull/86](https://github.com/JuliaStats/Distances.jl/pull/86)

src/Distances.jl
```diff
@@ -16,6 +16,8 @@ export
     pairwise!,
     evaluate,
 
+    fastview,
```
KristofferC: We probably don't want to export this

chethega: I think we should either export this, or a different way of accessing the allocation-free distance evaluation of matrix columns, or possibly both.

This was my original problem after all: I need to compute distances between more-or-less random columns in an inner loop (think ball-trees).

Would you feel better if I patched in bounds-checking for the fastview? Would be a no-op in inbounds contexts anyway...

Also, users who define their own distance functions get exposed to the fastview anyway (in their evaluate-function, if they call colwise or pairwise).

PR #88, opened 2017-11-29
Fix test_metricity failure to test

richardreeve:
- Fixes missing tests introduced by 3197ca5145dd6f44415c1b30bd4081fd454721ab
- Fixes F64 float conversion bug
- Fixes Base.Test deprecation for 0.7


KristofferC:
https://github.com/JuliaStats/Distances.jl/commit/3197ca5145dd6f44415c1b30bd4081fd454721ab did not miss any test. This is purely a display issue where tests in a function does not get added to the parent testset.

richardreeve:
No they aren't - now that I've added them in several fail. In fact I've now identified what has happened both here and in the pairwise tests. The tests were `quote`d from being a macro, so were never evaluated when the code was being used as a function. This is now fixed, but I am now fighting errors in the newly run tests.

richardreeve:
Okay. I think all of the test errors in 0.6 and 0.7 are now fixed. They were all problems either in F64 or in rounding errors in accumulations.

richardreeve:
@KristofferC In fact you can see there was a problem with PR #73 and that commit in particular because coverage dropped by 22.2%.

KristofferC:
Wow, that is amazing. Thanks for catching this.

KristofferC:
Isn't it enough to just remove the `quote`?

richardreeve:
Yes, I realised that a bit late :). I can go back if you prefer? That's what I did for test_pairwise(). However, I've also added in @testset statements to make it clearer where the errors are cropping up, and various other fixes.

richardreeve:
Okay, I've done that now. It hasn't made a massive difference to how intrusive it is I'm afraid, but it's more consistent with the other test groups anyway.

KristofferC:
Thanks, no point waiting for Mac worker. Coverage looks good again!

PR #90, opened 2018-02-19
Use a more performant formula for haversine

IlyaOrson:
```julia
Main> using BenchmarkTools

Main> a = rand()
0.8809105663628993

Main> @benchmark asin(sqrt($a))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     15.790 ns (0.00% GC)
  median time:      16.579 ns (0.00% GC)
  mean time:        17.691 ns (0.00% GC)
  maximum time:     712.929 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000

Main> @benchmark atan2(sqrt($a), sqrt(1-$a))
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
  minimum time:     24.080 ns (0.00% GC)
  median time:      24.870 ns (0.00% GC)
  mean time:        25.987 ns (0.00% GC)
  maximum time:     417.257 ns (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     1000
Main> asin(sqrt(a)) ≈ atan2(sqrt(a), sqrt(1-a))
true
```

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -33,7 +33,7 @@ function evaluate(dist::Haversine, x::VecOrLengthTwoTuple, y::VecOrLengthTwoTupl
     a = min(a, one(a))
 
     # distance on the sphere
-    2*dist.radius*atan2(√a, √(1-a))
+    2*dist.radius*asin(√a)
 end
 
 haversine(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple, radius::Real) = evaluate(Haversine(radius), x, y)
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -29,11 +29,8 @@ function evaluate(dist::Haversine, x::VecOrLengthTwoTuple, y::VecOrLengthTwoTupl
     # haversine formula
     a = sin(Δφ/2)^2 + cos(φ₁)*cos(φ₂)*sin(Δλ/2)^2
 
-    # take care of floating point errors
-    a = min(a, one(a))
-
     # distance on the sphere
-    2*dist.radius*asin(√a)
+    2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
 end
 
 haversine(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple, radius::Real) = evaluate(Haversine(radius), x, y)


juliohm:
@IlyaOrson it seems that the implementation with `asin` is more prone to floating point errors: http://www.movable-type.co.uk/scripts/latlong.html

`atan2` is used in most languages/libraries out there, even though the `asin` version is faster. If you can provide a good rounding error analysis, it would make it easier to assess the change.

IlyaOrson:
I think that the protection against rounding errors refers to `min(1, sqrt(a))` as discussed in the  [FAQ](https://www.movable-type.co.uk/scripts/gis-faq-5.1.html) of that link. This protection should be implicit in the existent code `a = min(a, one(a))`. 

[Wikipedia](https://en.wikipedia.org/wiki/Haversine_formula) suggest that both formulas suffer from rounding errors. If that is an issue, one should use the [Vincenty formula](https://en.wikipedia.org/wiki/Great-circle_distance#Computational_formulas) instead, maybe implemented separately as it is not the haversine distance after all. 

```julia
julia> myrange = [1 - big(10.0)^-i for i in 1:20];

julia> f(a) = asin(sqrt(a))

julia> g(a) = atan2(sqrt(a), sqrt(1-a))
g (generic function with 1 method)

julia> map(isapprox, f.(myrange), g.(myrange))
20-element Array{Bool,1}:
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
 true
```

juliohm:
Good point @IlyaOrson. Could you maybe update the PR to reflect the formula `2 ⋅ asin( min(1, √a) )` described in that link I shared? The current code is taking the minimum before the square root.

Then we could compare the speed and accuracy again before moving forward with the change.

juliohm:
Thank you @IlyaOrson, it is up to the Distances.jl maintainers now to assess the changes. If the precision is around the same as before, I like the performance improvement. :+1:

IlyaOrson:
Thanks!

PR #96, opened 2018-06-26
fix deprecations in Julia 0.7

heliosdrm:
Fixed deprecations observed in test for Jula 0.7.0-beta


heliosdrm:
Wait. Breaking change for 0.6, to be fixed.

KristofferC:
Personally, I think we can just upgrade to 0.7 now. Don't see much point releasing a new 0.6 release with only deprecation fixes for 0.7 in it.

REQUIRE
```diff
@@ -1,2 +1,2 @@
-julia 0.6
-Compat 0.62.0
+julia 0.7
+Compat 0.54.0
```
KristofferC: Can we just remove compat then?

heliosdrm:
Updated the PR to break 0.6 support and test only against 0.7 (+ nightly).

REQUIRE
heliosdrm: Yes, unless there is some other compat function or macro in the previous versions (I didn't check that).

KristofferC: I think this has to be `julia 0.7-` to show that it is for a prerelease of 0.7.

PR #107, opened 2018-08-27
Add periodic Euclidean metric

dkarrasch:
* add Euclidean metric on flat torus/cylinder
* updated README incl. benchmarks
* added tests

I had this code and felt encouraged to contribute by the discussion in [NearestNeighbors.jl#33](https://github.com/KristofferC/NearestNeighbors.jl/issues/33). I'm not sure if there are more periodic metrics of relevance and interest (squared Euclidean, weighted, ...), but these could be added as needed. I followed closely the code in `wmetrics.jl` which also has one parameter per dimension, but included the array-distinction as in `metrics.jl`.


KristofferC:
Im pretty sure this is possible to do without copy pasting much of the `evaluate` code,  by passing some extra state in the `s` variable.

dkarrasch:
Hm, I'm not sure I understand. The computational complexity (one paramter per dimension) is the same as in the weighted metrics case, and the `evaluate` implementation is exactly the same as the one there, just with overloads for `eval_op` etc. That's why one could call the evaluate instance from there, just that it would be confusing to call the `periods` field `weights`. That's why I felt there justification to copy and paste it. Could you please be more specific on how you think to change the implementation? If there is a cool trick by defining the `s` variable differently, this may be worth being transferred to the weighted metrics, I guess?

KristofferC:
The problem with copy pasting that whole loop block into multiple places is that when a performance fix like (https://github.com/JuliaStats/Distances.jl/commit/00452ab5f87ebc082d597198282b1d5964a41f27#diff-08d86a8ef0a417f2c2269c73f1dbfe5b) is done, it might not find all those copy pasted blocks. And indeed, that PR only changed one of them! We need to find a flexible enough way to write this loops so that it performs well and only exists in one place.

dkarrasch:
It took me a long while, but now it struck me, and I think I saw the light @KristofferC was sending in https://github.com/JuliaStats/Distances.jl/pull/107#issuecomment-419717358.

How about this design? Definition and constructors:
```
struct PeriodicEuclidean{N,T <: Real} <: Metric
    periods::NTuple{N,T}
end
PeriodicEuclidean(periods::Tuple) = PeriodicEuclidean(promote_type(typeof.(periods)...).(periods))
PeriodicEuclidean(periods::AbstractVector{T}) where {T<:Real} = PeriodicEuclidean(tuple(periods...))
```
Special methods:
```
result_type(dist::PeriodicEuclidean, ::AbstractArray{T1}, ::AbstractArray{T2}) where {T1, T2} =
    typeof(eval_end(dist, eval_op(dist, one(T1), one(T2))))
@inline function eval_start(d::PeriodicEuclidean, a::AbstractArray, b::AbstractArray)
    (zero(result_type(d, a, b)), d.periods)
end
@inline eval_op(::PeriodicEuclidean, ai, bi) = abs(ai - bi)
@inline function eval_reduce(::PeriodicEuclidean, s1, s2)
    li = first(s1[2])
    d = mod(s2, li)
    d = min(d, li - d)
    (s1[1] + abs2(d), Base.tail(s1[2]))
end
@inline eval_end(::PeriodicEuclidean, s) = sqrt(s[1])
```
As for the actual `evaluate` function, I use the verbatim `UnionMetrics`-instance. Does this look acceptable? If yes, then one could include the `UnionWeightedMetrics` into the classic `UnionMetrics`. One would have to find a way how to check the size compatibility of `a`, `b`, and the `periods`/`weights` fields in a generic `evaluate` function that would apply also to metrics that don't have a field.

If all this sounds good, then I would close this and start another PR "Unify all UnionMetrics".

dkarrasch:
I went ahead and implemented my suggestion just for the periodic/torus metric. Comments are very welcome. One question that I have is if `PeriodicEuclidean` is too long? What about `PEuclidean`, or something with torus? Even though this works also on cylinders with `Inf` entries in the periods-tuple.

dkarrasch:
Is there anything I can do to push this forward? This adds a new metric type and demonstrates how almost all metrics (i.e., `Union*Metrics`) can be brought under the hood of the same `evaluate` instance, in which then the loop can be optimized if necessary.

src/metrics.jl
```diff
@@ -140,6 +148,22 @@ see [`Euclidean`](@ref).
...
+"""
+    PeriodicEuclidean(L)
+Create a Euclidean metric on a rectangular periodic domain (i.e., a torus or
+a cylinder). Periods per dimension are contained in the vector `L`.
+For dimensions without periodicity put `Inf` in the respective component.
+
+# Example
+```
+julia> x, y, L = [0.0, 0.0], [0.7, 0.0], (0.5, Inf)
+([0.0, 0.0], [0.7, 0.0], (0.5, Inf))
```
nalimilan: Don't show this output using `;` in the line above.

src/metrics.jl
```diff
@@ -140,6 +148,22 @@ see [`Euclidean`](@ref).
...
+
+julia> evaluate(PeriodicEuclidean(L),x,y)
```
nalimilan: Show the output here. Also, add spaces after commas.

src/metrics.jl
```diff
@@ -140,6 +148,22 @@ see [`Euclidean`](@ref).
 """
 SqEuclidean() = SqEuclidean(0)
 
+"""
+    PeriodicEuclidean(L)
+Create a Euclidean metric on a rectangular periodic domain (i.e., a torus or
```
nalimilan: Add empty line.

Also, can you give the equation for this distance?

src/metrics.jl
```diff
@@ -140,6 +148,22 @@ see [`Euclidean`](@ref).
...
+
+# Example
+```
+julia> x, y, L = [0.0, 0.0], [0.7, 0.0], (0.5, Inf)
+([0.0, 0.0], [0.7, 0.0], (0.5, Inf))
+
+julia> evaluate(PeriodicEuclidean(L),x,y)
+```
+"""
+PeriodicEuclidean() = Euclidean()
```
nalimilan: It feels weird to have a constructor return an object of a different type. Better return a `PeriodicEuclidean` object, but define specialized methods for this special case (where `N==0`).

Also probably better define this at the same time as the `struct` (I'm not sure why `Euclidean` is defined here).

src/metrics.jl
```diff
@@ -99,8 +99,16 @@ struct MeanSqDeviation <: SemiMetric end
 struct RMSDeviation <: Metric end
 struct NormRMSDeviation <: Metric end
 
+struct PeriodicEuclidean{N,T <: Real} <: Metric
+    periods::NTuple{N,T}
+end
+PeriodicEuclidean(periods::Tuple) = PeriodicEuclidean(promote(periods...))
+PeriodicEuclidean(periods::AbstractVector{T}) where {T<:Real} = PeriodicEuclidean{length(periods),T}(tuple(periods...))
```
nalimilan: Why not require a tuple input?

src/metrics.jl
```diff
@@ -214,6 +250,28 @@ eval_end(::Euclidean, s) = sqrt(s)
...
+end
+@inline eval_op(::PeriodicEuclidean, ai, bi) = abs(ai - bi)
+@inline function eval_reduce(::PeriodicEuclidean, s1, s2)
+    li = first(s1[2])
+    d = mod(s2, li)
+    d = min(d, li - d)
+    (s1[1] + abs2(d), Base.tail(s1[2]))
+end
+@inline eval_end(::PeriodicEuclidean, s) = sqrt(s[1])
+function evaluate(dist::PeriodicEuclidean, a::T, b::T) where {T <: Number}
```
nalimilan: `Real` was used above instead of `Number`. Any reason to diverge?

dkarrasch: No, just oversight.

src/metrics.jl
dkarrasch: The idea here was to be liberal at construction, and make sure a tuple is constructed internally. Also, I had in mind adopting this approach to the weighted metrics. There, you can pass vectors of weights, and adopting the approach of this PR (with insisting on tuples) would be breaking for the weighted metrics.

dkarrasch:
@nalimilan Many thanks for the review. I'll make the requested changes and give a signal once I'm done. The only "controversial" aspect seems to be how liberal we want to be with passing a vector/tuple of periods at construction of the `PeriodicEuclidean` metric type, with a view of adopting the same approach to weighted `UnionMetrics`.

src/metrics.jl
nalimilan: AFAICT the situation is quite different for periods and for weights: the number of weights varies with the number of observations. OTC, the number of periods isn't supposed to vary much I guess, since it is a type parameter. But maybe that parameter should be dropped and periods should be stored as a vector instead? That would be consistent with the decision not to merge https://github.com/JuliaStats/Distances.jl/pull/115.

dkarrasch: This is exactly how I started out. Copying the weighted metric code, making the minimal modifications necessary for the periodic metric, and let go. That's when @KristofferC pointed out that code duplication is not good, and was pointing to an already existing issue with it (optimization of the for-loop in `evaluate`). But in fact, that same "criticism" could have been made earlier for the weighted metrics. Note that (regardless of the implementation) the weighted metrics and the periodic metric are of the exact same type: you have one parameter per observation. So @KristofferC 's suggestion was to somehow "hide" the dimensionwise parameter in an additional state in the `s` variable, and otherwise use the standard `evaluate` instance. The current state of this PR contains my take at this approach, and I do think that tuples are the right thing to use for type inference via `Base.tail` etc. The "fixed length" tuple thing is just due to my ignorance of a better type for this, but I just noticed that there is something like `Vararg{T}`, and I'll see if I can get this running.

nalimilan: OK, if the length of the periods is the same as the number of observations, then making it a type parameter is probably not appropriate. Better use a vector than a `Vararg`, though.

dkarrasch: But then the whole evaluate construction breaks down. It would force me to use indexing instead, which allocates...?

nalimilan: Can't you just pass the index where to start instead of calling `Base.tail(s1[2])` or its vector equivalent `s1[2][2:end]`?

dkarrasch: I'm afraid I can't pass an index because there is no such interface in the `UnionMetrics` `evaluate`, which I'm supposed (and want) to use. The vector equivalent allocates and doubles the runtime in the small example in the docstring, in comparison to my `NTuple` version.

Generally, what's the issue with having an `NTuple` for the parameters?

dkarrasch: Nevermind, the example was too specific. For the `NTuple` version, there was a significant slow-down for dimension 3 and higher. The slowdown is by far not as dramatic for `Vector`s. Unfortunately, for `Vector`s there are as many allocations as dimensions. I'll investigate further.

dkarrasch: Assuming that we're aiming at bringing all `UnionMetrics` under one `evaluate` umbrella, we could change the signature for `eval_op` throughout as follows:
```
eval_op(::Metric, ai, bi, i::Int=1)
```
As for `PeriodicEuclidean`, this yields the fastest implementation I could ever achieve. One would have to change all other `eval_op`s accordingly, most of which would not use it. This would also allow to "merge" the weighted metrics with the unweighted ones.

nalimilan: And what happens if you use `Iterators.drop(s1[2], 1)`?

dkarrasch: That seems to allocate and yield slower computations. I kind of like the above `eval_op` idea, since it mimicks what we have for the weighted metrics, it would not be breaking existing code, and it shouldn't slow down the classic metrics, should it?

nalimilan: Sounds like a good solution (though I don't master the whole machinery of this package).

dkarrasch: Unfortunately, this additional argument with default value seems to affect the classical metrics performancewise, which would not use it. Unless there are other ideas, I would suggest to merge the weighted metrics and the periodic metric here in one file, and declare them as (something like) `ParametricMetrics`. These would have their own `evaluate` instance and an additional parameter `i::Int=1` in their respective `eval_op`. This would also form a generic framework for metrics that have dimensionwise parameters. I'll do the changes and ask for comments.

nalimilan: How can it have an effect on performance if `i` isn't used? :-/

dkarrasch: I think it is because when `eval_op` is called with only 3 arguments `(::Metric, a, b)`, then it calls internally the 4-args instance and appends the default value `1`, which needs to be created. By "not used" I mean that it's value is not used in the function body, but the call routines still change.

nalimilan: All of this should happen at compile time and be optimized out by the compiler (at least with `@inline`). Can you show the code?

dkarrasch:
I hope I haven't messed up with all the commits on master and here. The current state is that the current dimension index `I` is added as an argument to `eval_op`. For the weighted metrics, this would be to replace the `wi`, and would give the opportunity to allow for more general dimensionwise parameters, in general. This should be checked, however, for performance regressions in the parameterfree, classic metrics. On my machine, I seem to measure a regression of about 10%. If this regression turns out to be reproducible and unavoidable, then we may need to have two sets of metrics: the parameter-free ones as in `metrics.jl`, and the parametric ones as in `wmetrics.jl` including the periodic one (I'm ignoring the other individual ones here). In that case, I would adopt (i) the loop and (ii) the two `evaluate` instances in `metrics.jl` to the `wmetrics.jl` case.

src/metrics.jl
```diff
@@ -180,70 +209,97 @@ end
             @simd for I in eachindex(a, b)
                 ai = a[I]
                 bi = b[I]
-                s = eval_reduce(d, s, eval_op(d, ai, bi))
+                s = eval_reduce(d, s, eval_op(d, ai, bi, I))
```
nalimilan: Actually, seeing the code I understand better what you're doing here. How about defining a `parameters(d)` method instead which would return the parameter vector? A fallback method would return `nothing` for most distances, and in that case you'd call the two-argument `eval_op`. That way indexing is clearly separated from actually computing the distance, most distances don't need any changes and you don't need to introduce the `ParametrizedMetrics` union.

dkarrasch: Sounds good, I'll give it a try.

dkarrasch:
@nalimilan It seems to work very well. I couldn't detect any regressions for the `evaluate(Euclidean(), a, b)`, all tests pass, so this should be good. I assume that the `p === nothing` branching gets compiled away.

I guess I would "merge" the weighted metrics into this framework in a different PR?

src/metrics.jl
```diff
@@ -99,8 +99,13 @@ struct MeanSqDeviation <: SemiMetric end
 struct RMSDeviation <: Metric end
 struct NormRMSDeviation <: Metric end
 
+const RealAbstractArray{T <: Real} =  AbstractArray{T}
 
-const UnionMetrics = Union{Euclidean,SqEuclidean,Chebyshev,Cityblock,TotalVariation,Minkowski,Hamming,Jaccard,RogersTanimoto,CosineDist,CorrDist,ChiSqDist,KLDivergence,RenyiDivergence,BrayCurtis,JSDivergence,SpanNormDist,GenKLDivergence}
+struct PeriodicEuclidean{W <: RealAbstractArray} <: Metric
```
nalimilan:
```suggestion
struct PeriodicEuclidean{W <: AbstractArray{<: Real}} <: Metric
```
Then you can drop `RealAbstractArray` above.

src/metrics.jl
```diff
@@ -180,20 +215,20 @@ end
...
                 bi = b[I]
-                s = eval_reduce(d, s, eval_op(d, ai, bi))
+                s = eval_reduce(d, s, p === nothing ? eval_op(d, ai, bi) : eval_op(d, ai, bi, p[I]))
             end
         else
             for (Ia, Ib) in zip(eachindex(a), eachindex(b))
                 ai = a[Ia]
                 bi = b[Ib]
-                s = eval_reduce(d, s, eval_op(d, ai, bi))
+                s = eval_reduce(d, s, p === nothing ? eval_op(d, ai, bi) : eval_op(d, ai, bi, p[Ia]))
```
nalimilan: For correctness I think you'll have to write a separate loop with `(Ia, Ib, Ip) in zip(eachindex(a), eachindex(b), eachindex(p))`. Indeed, let's imagine `a` is 3×2 and `b` is 2×3 then `p` may be 2×3 and if `Base.IndexStyle(a) == CartesianIndexing()` we'll try to call `p[3, 2]` which won't work (I know, that's a contrived example, but...).

src/metrics.jl
```diff
@@ -180,20 +215,20 @@ end
...
             end
         end
     end
     return eval_end(d, s)
 end
 result_type(dist::UnionMetrics, ::AbstractArray{T1}, ::AbstractArray{T2}) where {T1, T2} =
-    typeof(eval_end(dist, eval_op(dist, one(T1), one(T2))))
+    typeof(parameters(dist) === nothing ? eval_end(dist, eval_op(dist, one(T1), one(T2))) : evaluate(dist, one(T1), one(T2)))
```
nalimilan: Why not call `eval_end` where there are parameters? Also, add a line break after `:`.

src/metrics.jl
```diff
@@ -214,6 +249,29 @@ eval_end(::Euclidean, s) = sqrt(s)
...
 
+# PeriodicEuclidean
+Base.eltype(d::PeriodicEuclidean) = eltype(d.periods)
+@inline parameters(d::PeriodicEuclidean) = d.periods
+@inline function eval_start(d::PeriodicEuclidean, a::AbstractArray, b::AbstractArray)
+    zero(result_type(d, a, b))
+end
+@inline function eval_op(d::PeriodicEuclidean, ai, bi, pi)
+    s = abs(ai - bi)
+    s = mod(s, pi)
```
nalimilan: Are we sure `s` won't change type here? In doubt, maybe better use different names for the variable (same below).

src/metrics.jl
```diff
@@ -214,6 +249,29 @@ eval_end(::Euclidean, s) = sqrt(s)
...
+    abs2(s)
+end
+@inline eval_reduce(::PeriodicEuclidean, s1, s2) = s1 + s2
+@inline eval_end(::PeriodicEuclidean, s) = sqrt(s)
+function evaluate(dist::PeriodicEuclidean, a::T, b::T) where {T <: Real}
+    p = isempty(dist.periods) ? one(eltype(dist)) : first(dist.periods)
+    d = mod(abs(a - b), p)
+    d = min(d, p - d)
+end
+peuclidean(a::AbstractArray, b::AbstractArray, p) = evaluate(PeriodicEuclidean(p), a, b)
```
nalimilan:
```suggestion
peuclidean(a::AbstractArray, b::AbstractArray, p::AbstractArray{<: Real}) =
    evaluate(PeriodicEuclidean(p), a, b)
```

README.md
```diff
@@ -212,29 +214,30 @@ The table below compares the performance (measured in terms of average elapsed t
...
+| RenyiDivergence | 0.051003s |  0.045333s |  1.1251 |
+| JSDivergence | 0.046896s |  0.042365s |  1.1070 |
+| BhattacharyyaDist | 0.008461s |  0.003804s |  2.2244 |
+| HellingerDist | 0.008584s |  0.003678s |  2.3341 |
+| WeightedSqEuclidean | 0.005481s |  0.001182s |  4.6370 |
+| WeightedEuclidean | 0.005591s |  0.001198s |  4.6680 |
+| WeightedCityblock | 0.005712s |  0.001177s |  4.8521 |
+| WeightedMinkowski | 0.048870s |  0.041635s |  1.1738 |
+| WeightedHamming | 0.005682s |  0.001183s |  4.8042 |
+| PeriodicEuclidean | 0.012928s |  0.007914s |  1.6335 |
```
nalimilan: Put it with `Euclidean` for consistency.

README.md
```diff
@@ -166,7 +167,8 @@ Each distance corresponds to a distance type. The type name and the correspondin
 |  WeightedCityblock   |  `wcityblock(x, y, w)`     | `sum(abs(x - y) .* w)`  |
 |  WeightedMinkowski   |  `wminkowski(x, y, w, p)`  | `sum(abs(x - y).^p .* w) ^ (1/p)` |
 |  WeightedHamming     |  `whamming(x, y, w)`       | `sum((x .!= y) .* w)`  |
-|  Bregman             |  `bregman(F, ∇, x, y; inner = LinearAlgebra.dot)` | `F(x) - F(y) - inner(∇(y), x - y)` | 
+|  PeriodicEuclidean   |  `peuclidean(x, y, p)`     | `sqrt(sum(min(mod(abs(x - y), p), p - mod(abs(x - y), p)).^2))`  |
```
nalimilan: Put it with `Euclidean` for consistency.

src/metrics.jl
```diff
@@ -140,6 +145,26 @@ see [`Euclidean`](@ref).
...
+
+# Example
+```jldoctest
+julia> x, y, L = [0.0, 0.0], [0.75, 0.0], [0.5, Inf];
+
+julia> evaluate(PeriodicEuclidean(L), x, y)
+0.25
+```
+"""
+PeriodicEuclidean() = PeriodicEuclidean([])
```
nalimilan:
```suggestion
PeriodicEuclidean() = PeriodicEuclidean(Int[])
```

src/metrics.jl
```diff
@@ -214,6 +249,29 @@ eval_end(::Euclidean, s) = sqrt(s)
...
+@inline eval_reduce(::PeriodicEuclidean, s1, s2) = s1 + s2
+@inline eval_end(::PeriodicEuclidean, s) = sqrt(s)
+function evaluate(dist::PeriodicEuclidean, a::T, b::T) where {T <: Real}
+    p = isempty(dist.periods) ? one(eltype(dist)) : first(dist.periods)
+    d = mod(abs(a - b), p)
+    d = min(d, p - d)
+end
+peuclidean(a::AbstractArray, b::AbstractArray, p) = evaluate(PeriodicEuclidean(p), a, b)
+peuclidean(a::AbstractArray, b::AbstractArray) = euclidean(a, b)
+peuclidean(a::Real, b::Real, p::Real) = evaluate(PeriodicEuclidean([p]), a, b)
```
nalimilan: Wouldn't `Number` be OK too?

nalimilan:
@KristofferC Are you OK with this approach?

src/metrics.jl
dkarrasch: Sure, I just copied it from the `wmetrics.jl`.

README.md
dkarrasch: Yes, I need to run the benchmarks again. This is still the from the early submission, when I tried to follow closely the weighted metrics code.

dkarrasch:
It remains to update the README. On my current machine, the method disambiguation for `CorrDist` doesn't work, so I'll try tomorrow at work. Interestingly, even tests fail here because of that whereas they don't on Travis... 🤔

dkarrasch:
I have now written out all branches explicitly to handle indexing correctly. Also, I changed back the for-loop to go over `eachindex(a, b)` or `eachindex(a, b, p)`, which results in some noticeable speed-up. I'm not getting why, for increased vector lengths, the ratio between `PeriodicEuclidean` and `Euclidean` runtimes is getting worse and worse, albeit the `PeriodicEuclidean` runtime scales roughly linearly with the number of dimensions. I guess the periodic calculation somehow does not benefit from `@simd`?

nalimilan:
Only `@code_llvm` or `@code_native` can tell...

dkarrasch:
Yes, unfortunately I don't know how to read/examine that. Maybe I should try to bring one of the weighted metrics into this periodic format, and compare them, to see if there are issues with the general "architecture" implemented here, or if it's due to the involved extra operations like `mod`.

nalimilan:
Note that `Euclidean` is probably very fast since IIRC it uses BLAS tricks. Typically, calling BLAS has an overhead for small arrays, so that could explain the widening gap.

dkarrasch:
That's true, but I think it uses BLAS tricks only for the `pairwise` case. I was only comparing the computation of distances of two longer and longer vectors. I don't think that the periodic distance would be used with extensively long vectors (many observations), but there could be an issue to be investigated. Let's see what @KristofferC has to say.

KristofferC:
Sorry been a bit AFK w.r.t this package. Will look over this now

KristofferC:
I see a 2x regression for

```
using Distances
using BenchmarkTools

m = 10
n=100_000
Xcol = rand(m,n)
Ycol=rand(m,n)


function colwise_naive(X,Y)
    m = size(X,1)
    n = size(X,2)
    T = eltype(X)
    @assert (m == size(Y,1) && n == size(Y,2) && T==eltype(Y))

    res = fill(zero(eltype(X)), n)
    @inbounds  for j = 1:n
        s_ = zero(T)
        @inbounds @simd for i = 1:m
            dd_ = X[i,j]-Y[i,j]
            s_ += dd_*dd_
        end
        res[j] = sqrt(s_)
    end
    res
end

dd = Euclidean()
@btime colwise(dd, Xcol,Ycol)
@btime colwise_naive(Xcol,Ycol)
```

on this branch. On master these are both equally fast.

dkarrasch:
Yes, I can confirm. 😞

I would leave the standard metrics in peace and try to gently rewrite the weighted metrics in a way that passes the index `i` instead of the weight `wi` to `eval_op`, in order to allow for more general "parametric" metrics? Does that sound acceptable (if I manage to do so without regression)? That would at least keep the numer of `evaluate` instances constant while adding a metric.

nalimilan:
Do you have an idea where the slowdown comes from exactly? AFAICT the only changes in this PR which could affect `Euclidean` are the `p === nothing` branches, which should be eliminated at compile time. I'd check whether that's the case and whether allocations happen at unexpected places.

dkarrasch:
Yes, there seems to be an allocations issue, but I don't understand why. When I test a single `evaluate`, then there are no allocations.
```
@btime evaluate($dd, $(Xcol[:,1]), $(Ycol[:,1]))
  11.421 ns (0 allocations: 0 bytes)
0.8188840435814232
```

nalimilan:
Can you see using `--track-allocations=user` at which lines they appear? Does `@code_warntype` reveal a type unstability?

KristofferC:
Last time I looked at this I filed https://github.com/JuliaLang/julia/issues/28226

nalimilan:
OK, so maybe that's just due to using `eachindex` instead of `1:n`?

dkarrasch:
I'm so sorry for the noise. I did extensive benchmarks, but only with single `evaluate` calls. Based on these tests, I went back to the `eachindex` looping, because that seemed to be faster. So, yes, the regression is gone when keeping the `1:length(a)` loop.

dkarrasch:
I skrewed up with git, so I'll start with the current master from scratch.

PR #115, opened 2018-11-26
make Minkowski distance a parametric type

### LINKED ISSUE #56: Consider making Minkowski distance a parametric type
It would be nice to be able to dispatch on the `p` parameter in the `Minkowski` distance type. There is a very similar implementation here (thanks to @Evizero), which I think could be ported here if others are interested in this functionality: https://github.com/JuliaML/Losses.jl/blob/master/src/supervised/distance.jl#L10

dkarrasch:
This fixes #56. Unfortunately, this (currently) does not yield any performance improvements except for `p=1` and `p=2` (which is nothing but `Euclidean()`). At least, it doesn't seem to make things worse. Comments welcome.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -17,9 +17,8 @@ struct Cityblock <: Metric end
 struct Jaccard <: Metric end
 struct RogersTanimoto <: Metric end
 
-struct Minkowski{T <: Real} <: Metric
-    p::T
-end
+struct Minkowski{p} <: Metric end
+Minkowski(p::T) where {T <: Real} = Minkowski{p}()
 
 struct Hamming <: Metric end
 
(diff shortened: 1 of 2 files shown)


dkarrasch:
Is this change still of interest? If yes, could this be merged? Or is there anything that I should add?

dkarrasch:
I guess this is potentially code-breaking, since for `m = Minkowski(2.5)` there is no `m.p`. Is there any way to provide a work-around here? It's perhaps not as important. Otherwise, the constructor and the `evaluate` calls are functionally identical to the previous implementation. It would be great to come to a conclusion regarding this issue.

KristofferC:
Unless there is a noticeable performance difference, I don't see much reason to do this. It will cause more things to have to be compiled, might break some code and it will severely degrade performance for those that currently store such a distance in a struct (since after this PR it will miss the type parameter).

PR #116, opened 2018-12-10
Get docstring without @doc to avoid error when compiling

### LINKED ISSUE #113: Doc macro makes compilation fail
I'm trying to compile some packages (using ApplicationBuilder/PackageCompiler) and this line makes the process fail:

https://github.com/JuliaStats/Distances.jl/blob/master/src/metrics.jl#L328

```julia
ERROR: LoadError: LoadError: LoadError: LoadError: LoadError: LoadError: LoadError: LoadError: LoadError: LoadError: UndefRefError: access to undefined reference
Stacktrace:
 [1] top-level scope at none:0 (repeats 4 times)
in expression starting at /Users/bieler/.julia/packages/Distances/nLAdT/src/metrics.jl:328
```

Is there another way to do this doc merging ?

donm:
Fixes #113.

I'm not exactly sure what the problem was, because the error was happening during compilation with PackageCompiler but not when simply loading the module.  In any case, using `@doc` for getting the docstring led to trying to look up docstring at parse time in a context where `RenyiDivergence` was undefined, and this change shifts the lookup to happen during evaluation of the module definition.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -332,10 +332,12 @@ function eval_end(dist::RenyiDivergence, s::Tuple{T,T,T,T}) where {T <: Real}
     end
 end
 
-# Combine docs with RenyiDivergence
-@doc (@doc RenyiDivergence) renyi_divergence
-
 renyi_divergence(a::AbstractArray, b::AbstractArray, q::Real) = evaluate(RenyiDivergence(q), a, b)
+# Combine docs with RenyiDivergence. Getting the docstring with @doc causes
+# problems with package compilation.
+let docstring = Base.Docs.getdoc(RenyiDivergence)
+    @doc docstring renyi_divergence
+end
 
 # JSDivergence
 @inline function eval_op(::JSDivergence, ai::T, bi::T) where {T}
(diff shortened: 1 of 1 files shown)


zgornel:
Is someone handling this pull ? Seems really strange that the coverage decreased by 39% ... This should get merged as it solves the compilation issue. Version should get bumped as well.
@donm Any idea on the coveralls fail ?

dkarrasch:
Coveralls "fails" only because coverage decreased, I think that's the criterion. And I think coverage decreased because of an update in the coverage software, or something like this. I had the same issue in a different package, and the same happened in #115.

KristofferC:
I dont think this is an Documenter issue? The `@doc` macro is in base and nothing suggested anything else having to do with Documenter here? Or am I misunderstanding something?

nalimilan:
Right, in Base then. :-)

donm:
I opened an issue in the Julia repo here: https://github.com/JuliaLang/julia/issues/31640

In the meantime, do we want to go ahead and merge this or wait?

nalimilan:
I guess we have to merge this since it won't be fixed in Base before 1.2.1 or 1.3. Can you add a link to the Julia issue in the comment?

PR #118, opened 2018-12-11
Total variation distance

devmotion:
Hi!

I noticed that the [total variation distance](https://en.wikipedia.org/wiki/Total_variation_distance_of_probability_measures) is not included in this package. However, although it is one of the most prominent [statistical distances](https://en.wikipedia.org/wiki/Statistical_distance#Examples) I'm not sure whether it's actually reasonable to add it since it's proportional to the Cityblock distance which is already included.

I updated the tests and benchmarks in case you think it should be added nevertheless.


devmotion:
Any comments or opinions?

KristofferC:
Looks good to me. As you say, it is a bit redundant with CityBlock but the amount of code is small and we might as well include it here and not make everyone write the wrapper themselves.

KristofferC:
Close + reopen to restart CI

devmotion:
Thanks for taking care of this PR!

Issue #35, opened 2015-12-11
pairwise to work on rows instead of columns.


musm:
Say I have a large matrix X in memory
I would like to use `pairwise` operating on the rows instead of the columns, without having to pass the transpose of X (so to not allocate another object in memory).

rawls238:
I think this is a good idea, I'll put up an implementation of this unless @KristofferC has any objections about adding this

musmo:
What would be a minimally invasive approach?

nalimilan:
`transpose` now returns a lazy wrapper, so this may no longer be an issue.

OTOH it could be clearer for users to be able to specify the dimension over which variables are stored, since we differ from the convention used by other languages (see also https://github.com/JuliaStats/Clustering.jl/issues/79). Another reason is that `cov` supports such an argument (and expects variables to be stored by column by default...).

nalimilan:
I've filed https://github.com/JuliaStats/Distances.jl/pull/121 taking advantage of `transpose` to support a `dims` argument. Please comment there.

PR #121, opened 2019-02-09
Add dims argument to pairwise

nalimilan:
This allows specifying whether observations are stored as columns (`dims=2`) or as rows (`dims=1`).
All functions are still written internally to perform computations over columns, computations over rows are supported by calling transpose on the matrices.

Add a deprecation to require specifying explicitly `dims=1` or `dims=2`: contrary to this package, other implementations most frequently default to `dims=1`, and even in Julia the `cov` and `cor` functions default to `dims=1`.

Fixes https://github.com/JuliaStats/Distances.jl/issues/35.

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -92,10 +92,8 @@ end
 ###########################################################
 
 function sqrt!(a::AbstractArray)
-    @inbounds @simd for i in eachindex(a)
-        x = a[i]
-        # > 0 is there to tolerate precision issues
-        a[i] = x > 0 ? sqrt(x) : zero(x)
+    @simd for i in eachindex(a)
+        @inbounds a[i] = sqrt(a[i])
     end
     a
 end
--- a/src/common.jl
+++ b/src/common.jl
@@ -35,7 +35,7 @@ end
 function get_pairwise_dims(r::AbstractMatrix, a::AbstractMatrix, b::AbstractMatrix)
     ma, na = size(a)
     mb, nb = size(b)
-    ma == mb || throw(DimensionMismatch("The numbers of rows in a and b must match."))
+    ma == mb || throw(DimensionMismatch("The numbers of rows or columns in a and b must match."))
     size(r) == (na, nb) || throw(DimensionMismatch("Incorrect size of r."))
     return (ma, na, nb)
 end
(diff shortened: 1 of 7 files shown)


src/common.jl
```diff
@@ -92,8 +92,10 @@ end
 ###########################################################
 
 function sqrt!(a::AbstractArray)
-    @simd for i in eachindex(a)
-        @inbounds a[i] = sqrt(a[i])
+    @inbounds @simd for i in eachindex(a)
+        x = a[i]
+        # > 0 is there to tolerate precision issues
+        a[i] = x > 0 ? sqrt(x) : zero(x)
```
nalimilan: I really don't understand why this is needed after my PR but not before it. Without adding this check I get errors because the Mahalanobis distance tries to compute `sqrt(-3.552713678800501e-15)`. This looks like an acceptable approximation to me, but how can my PR have triggered this? Also for safety it would make sense to check that `x` remains small, but without hurting performance too much.

andreasnoack: I don't think this is the right place to add this. An in-place square root shouldn't truncate here but throw for negative values and the branch will also break vectorization.

nalimilan: As I said I don't know why my PR makes this necessary, but without it tests don't pass.

Regarding vectorization, LLVM is able to eliminate the branch and generate SIMD instructions. So this makes absolutely no difference to performance:
```julia
julia> function f(a)
           @inbounds @simd for i in eachindex(a)
               x = a[i]
               a[i] = x > 0 ? sqrt(x) : zero(x)
           end
           a
       end

julia> function g(a)
           @inbounds @simd for i in eachindex(a)
               x = a[i]
               a[i] = sqrt(x)
           end
           a
       end
g (generic function with 1 method)

julia> x = rand(10_000);

julia> @btime f(x);
  20.076 μs (0 allocations: 0 bytes)

julia> @btime g(x);
  20.076 μs (0 allocations: 0 bytes)
```

Overall I think the question is really: is it normal that small negative values are produced? If so, then we need to be prepared to handle them. We could also round them to zero when computing distances. I'm not sure what the tolerance should be, but `eps(1.0)` isn't enough (`sqrt(eps(1.0))` is OK, though).

nalimilan: OK, I've found why the tests started failing: I had removed a special method for when a single matrix is passed, which takes advantage from symmetry, which apparently ensures the result is always positive. I'm not sure whether small negative values can legitimately happen, but that's out of the scope of this PR so I have reverted the `sqrt!` change.

nalimilan: Actually the problem of small negative values is already known: https://github.com/JuliaStats/Distances.jl/pull/13

PR #128, opened 2019-04-26
restrict eltype to Number

johnnychen94:
There're some cases we calculate distances between `a::AbstractArray{T}, b::AbstractArray{T}` where `T` is not a number. For example, in image-processing `where {T<:AbstractArray}`.

By change from `AbstractArray{T}` to `AbstractArray{<:Number}`, this PR provides the ability for downstream packages(e.g., `ImageDistances`) to reuse codes in `Distances` with a simple dispatch strategy

The test says ok, but IF there's some `Premetric` that is not limited to `AbstractArray{<:Number}` type, please point it out.

### DIFF
--- a/src/bregman.jl
+++ b/src/bregman.jl
@@ -22,7 +22,7 @@ end
 Bregman(F, ∇) =  Bregman(F, ∇, LinearAlgebra.dot)
 
 # Evaluation fuction 
-function evaluate(dist::Bregman, p::AbstractVector, q::AbstractVector)
+function evaluate(dist::Bregman, p::AbstractVector{<:Number}, q::AbstractVector{<:Number})
     # Create cache vals.
     FP_val = dist.F(p);
     FQ_val = dist.F(q); 
--- a/src/common.jl
+++ b/src/common.jl
@@ -6,33 +6,33 @@
 #
 ###########################################################
 
-function get_common_ncols(a::AbstractMatrix, b::AbstractMatrix)
+function get_common_ncols(a::AbstractMatrix{<:Number}, b::AbstractMatrix{<:Number})
     na = size(a, 2)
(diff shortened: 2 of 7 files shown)


nalimilan:
This is going to make it impossible to call these functions on arrays of `Number` objects whose eltype isn't `<:Number`, notably  `Union{<:Number, Missing}`. That goes against the work to support such arrays e.g. in StatsBase.

Can't you use a different strategy in ImageDistances? Note that defining methods defined in Distances for arguments of type `AbstractArray{<:AbstractArray}` in ImageDistances would technically be type piracy since ImageDistances doesn't own the function nor the type.

johnnychen94:
Thanks for the feedback, wasn’t aware of these concerns, I’ll rethink about it carefully to dealing with such issue in ImageDistances.

In the meantime, will it possible to change Number to “const NumberLike = Union{Number, Missing}”? IMO AbstractArray is still too generic, regardless of my issue.

Johnny
On Apr 29, 2019, 20:20 +0800, Milan Bouchet-Valat <notifications@github.com>, wrote:

This is going to make it impossible to call these functions on arrays of Number objects whose eltype isn't <:Number, notably Union{<:Number, Missing}. That goes against the work to support such arrays e.g. in StatsBase.

Can't you use a different strategy in ImageDistances? Note that defining methods defined in Distances for arguments of type AbstractArray{<:AbstractArray} in ImageDistances would technically be type piracy since ImageDistances doesn't own the function nor the type.

nalimilan:
> In the meantime, will it possible to change `Number` to `const NumberLike = Union{Number, Missing}`? IMO `AbstractArray` is still too generic, regardless of my issue.

In general I think we try to keep types as broad as possible, just like you are allowed to index into an array using a `Vector{Any}` in Base.

johnnychen94:
Fixed in ImageDistances.jl side

PR #132, opened 2019-05-02
add AbstractArray constraint to deviations

johnnychen94:
All other `evaluate` and helper functions in `Distances` have this constraint.

Recently I'm implementing a downstream package `ImageDistances` based on `Distances`, adding this constraint in `Distances` make it much easier to handle dispatches.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -442,20 +442,20 @@ rogerstanimoto(a::AbstractArray{T}, b::AbstractArray{T}) where {T <: Bool} = eva
 
 # Deviations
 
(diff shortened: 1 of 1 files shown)


nalimilan:
AFAICT other `evaluate` methods also accept `Number`s, right?

johnnychen94:
> AFAICT other evaluate methods also accept Numbers, right?
https://github.com/JuliaStats/Distances.jl/blob/bb54e3b425d5804d4b5894e2d619a630e20c6873/src/metrics.jl#L201

nalimilan:
But deviations are not in `UnionMetrics`.

Issue #130, opened 2019-04-26
Compatibility WeightedSqEuclidean with Tracker


theogf:
When using `pairwise()` with a `WeightedSqEuclidean` metric, it is not possible to use Tracker.jl.
When computing the gradient the [`wsumsq_percol()`](https://github.com/JuliaStats/Distances.jl/blob/bb54e3b425d5804d4b5894e2d619a630e20c6873/src/common.jl#L112) function is problematic : 

```julia
function wsumsq_percol(w::AbstractArray{T1}, a::AbstractMatrix{T2}) where {T1, T2}
    m = size(a, 1)
    n = size(a, 2)
    T = typeof(one(T1) * one(T2))
    r = Vector{T}(undef, n)
    for j = 1:n
        aj = view(a, :, j)
        s = zero(T)
        @simd for i = 1:m
            @inbounds s += w[i] * abs2(aj[i])
        end
        r[j] = s
    end
    return r
end
```
`T1` will not be of type `TrackedReal{Float}` but `Float`, therefore `r` will be also `Float`. So the allocation `r[j] = s` fails since `s` is of type `TrackedReal`.
Changing `one(T1)` by `one(w[1])` solves the problem but is obviously not the best answer to the problem.

nalimilan:
Maybe the loop could be switched to use an array comprehension instead?

PR #134, opened 2019-05-27
Loosen type annotations and fix https://github.com/JuliaStats/Distances.jl/issues/130

devmotion:
When I started working on this PR, my intention was to fix https://github.com/JuliaStats/Distances.jl/issues/130. Actually only the changes in https://github.com/devmotion/Distances.jl/blob/promote_op/src/common.jl#L115 and https://github.com/devmotion/Distances.jl/blob/promote_op/src/wmetrics.jl#L46 are needed to be able to run successfully
```julia
using Distances, Tracker

Tracker.gradient(x -> sum(pairwise(WeightedSqEuclidean(x), [3 5; 4 6]; dims = 2)), [1, 2])

y, back = Tracker.forward(x -> sum(pairwise(WeightedSqEuclidean([1,2]), x; dims = 2)), [3 5; 4 6])
back(1)
```
BTW, I don't know why but trying to execute
```julia
Tracker.gradient(x -> sum(pairwise(WeightedSqEuclidean([1, 2]), x; dims = 2)), [3 5; 4 6])
```
yields the error message
```julia
ERROR: MethodError: Cannot `convert` an object of type Array{Float64,1} to an object of type SubArray{Float64,1,Array{Float64,2},Tuple{Base.Slice{Base.OneTo{Int64}},Int64},true}                                           
Closest candidates are:
  convert(::Type{T<:AbstractArray}, ::T<:AbstractArray) where T<:AbstractArray at abstractarray.jl:14
  convert(::Type{T<:AbstractArray}, ::LinearAlgebra.Factorization) where T<:AbstractArray at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.1/LinearAlgebra/src/factorization.jl:46
  convert(::Type{T}, ::T) where T at essentials.jl:154
  ...
```
However, I noticed that due to the quite restrictive type annotations for scalar inputs not only typically scalar evaluation is restricted to inputs of exactly the same type, but for weighted metrics also the vectorized computations are restricted to arguments with elements of exactly the same type. After the simple change to `promote_op` in https://github.com/devmotion/Distances.jl/blob/promote_op/src/wmetrics.jl#L46, this behaviour is still present since in all cases with arguments of different types `result_type` would return `Any`, and hence evaluation breaks as soon as `zero(result_type(...))` is called. Thus I loosened the type annotations of the scalar case to fix this issue.

Moreover, I applied the same changes to the standard metrics.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -252,22 +252,21 @@ end
     end
     return eval_end(d, s)
 end
-result_type(dist::UnionMetrics, ::AbstractArray{T1}, ::AbstractArray{T2}) where {T1, T2} =
-    typeof(eval_end(dist, parameters(dist) === nothing ?
-                        eval_op(dist, one(T1), one(T2)) :
-                        eval_op(dist, one(T1), one(T2), one(eltype(dist)))))
+result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray) =
+    Base.promote_op(evaluate, typeof(dist), eltype(a), eltype(b))
+
 eval_start(d::UnionMetrics, a::AbstractArray, b::AbstractArray) =
     zero(result_type(d, a, b))
 eval_end(d::UnionMetrics, s) = s
 
-evaluate(dist::UnionMetrics, a::T, b::T) where {T <: Number} = eval_end(dist, eval_op(dist, a, b))
+evaluate(dist::UnionMetrics, a::Number, b::Number) = eval_end(dist, eval_op(dist, a, b))
 
 # SqEuclidean
 @inline eval_op(::SqEuclidean, ai, bi) = abs2(ai - bi)
 @inline eval_reduce(::SqEuclidean, s1, s2) = s1 + s2
 
 sqeuclidean(a::AbstractArray, b::AbstractArray) = evaluate(SqEuclidean(), a, b)
-sqeuclidean(a::T, b::T) where {T <: Number} = evaluate(SqEuclidean(), a, b)
+sqeuclidean(a::Number, b::Number) = evaluate(SqEuclidean(), a, b)
 
 # Euclidean
 @inline eval_op(::Euclidean, ai, bi) = abs2(ai - bi)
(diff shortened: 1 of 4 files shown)


ararslan:
`promote_op` is not a great thing to use unless it can't be avoided at all, as in the case of determining the element type for an empty collection based on one or more other empty collections. (I believe the docstring for it actually makes a note to this effect.)

devmotion:
Ah OK, good to know! I had not checked the docstring, from just reading the documentation https://docs.julialang.org/en/v1.1/manual/methods/#Output-type-computation-1 I had got the impression `promote_op` is actually the preferred way to perform these output computations.

ararslan:
From the docs:
```julia
    # this is wrong, since depending on the return value
    # of type-inference is very brittle (as well as not being optimizable):
    # R = Base.return_types(op, (eltype(a), eltype(b)))

    ## but, finally, this works:
    R = promote_op(op, eltype(a), eltype(b))
    ## although sometimes it may give a larger type than desired
    ## it will always give a correct type
```
This is interesting in that it's inaccurate. As of Julia 1.1, `promote_op` is just a wrapper over getting the return type from inference, so it's equivalent to the one that's claimed to be brittle. It's also not always a correct type; I ran into that last week.

ararslan:
This is looking good, thanks! Would you be willing to add some tests for mismatched input types that would go through these loosened signatures?

ararslan:
Wow, you _seriously_ came through with the tests. Nice job! :smile:

src/metrics.jl
```diff
@@ -285,13 +284,13 @@ Base.eltype(d::PeriodicEuclidean) = eltype(d.periods)
     s3 = min(s2, p - s2)
     abs2(s3)
 end
+@inline function eval_op(d::PeriodicEuclidean, ai, bi)
+    periods = d.periods
+    p = isempty(periods) ? oneunit(eltype(periods)) : first(periods)
```
nalimilan: Why is `oneunit(...)` appropriate here when the periods are empty? Is this tested?

dkarrasch: I'm not sure this function gets ever called? The empty test should be tested, I remember that I was struggling with that case at some point. I think, however, that in the empty periods case, `evaluate` will have checked before that `a` and `b` have also length zero, and that it doesn't make any difference what you put here for `p`. Apparently, I was doing something similar in the previous line 258 above.

Issue #126, opened 2019-04-20
Factor Out Common Distance API to Leverage `eval_op`, ... etc.


bhgomes:
I think it would be beneficial to factor out the `eval_op`, `eval_end`, `eval_reduce` API so that others who want to define custom metrics (like #95) can extend the already optimized `evaluate` that is built off of them. Not sure how this would interact with `UnionMetrics`.

dkarrasch:
@bhgomes Could it be that this is resolved by #163? Depending on what you want to define, `PreMetric`, `SemiMetric` or `Metric`, you can subtype your custom metric to `Union[Pre/Semi]Metric`, overload those functions that are not covered by the generic fallbacks, and otherwise let the package do its job. See https://github.com/JuliaStats/Distances.jl/issues/151#issuecomment-612460855 for an example.

dkarrasch:
Closed by #163.

Issue #100, opened 2018-07-22
Regression of colwise on Julia 0.7?


dkarrasch:
I was testing some code of mine in Julia 0.7, and found a rather drastic regression.

On Julia 0.7:
```
julia> using Distances, BenchmarkTools

julia> A = rand(2,41); B = rand(2,41);

julia> @benchmark colwise($(Euclidean()), $A, $B)
BenchmarkTools.Trial: 
  memory estimate:  4.28 KiB
  allocs estimate:  83
  --------------
  minimum time:     634.094 ns (0.00% GC)
  median time:      671.082 ns (0.00% GC)
  mean time:        786.797 ns (12.98% GC)
  maximum time:     176.285 μs (99.54% GC)
  --------------
  samples:          10000
  evals/sample:     170

julia> d = fill(0., 41);

julia> @benchmark colwise!($d,$(Euclidean()), $A, $B)
BenchmarkTools.Trial: 
  memory estimate:  3.84 KiB
  allocs estimate:  82
  --------------
  minimum time:     567.255 ns (0.00% GC)
  median time:      610.462 ns (0.00% GC)
  mean time:        753.263 ns (16.55% GC)
  maximum time:     175.393 μs (99.61% GC)
  --------------
  samples:          10000
  evals/sample:     184

julia> versioninfo()
Julia Version 0.7.0-beta2.0
Commit b145832402* (2018-07-13 19:54 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin14.5.0)
  CPU: Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.0 (ORCJIT, skylake)
```
(note the allocations even for `colwise!`), and for comparison
```
julia> @benchmark colwise($(Euclidean()), $A, $B)
BenchmarkTools.Trial:
  memory estimate:  448 bytes
  allocs estimate:  1
  --------------
  minimum time:     246.176 ns (0.00% GC)
  median time:      251.361 ns (0.00% GC)
  mean time:        278.406 ns (2.41% GC)
  maximum time:     2.926 μs (0.00% GC)
  --------------
  samples:          10000
  evals/sample:     386

julia> d = zeros(41);

julia> @benchmark colwise!($d,$(Euclidean()), $A, $B)
BenchmarkTools.Trial:
  memory estimate:  0 bytes
  allocs estimate:  0
  --------------
[...]

Issue #95, opened 2018-06-20
Is there a way for the user to define a metric by themself?


000Justin000:

dkarrasch:
Sure, all metrics in the package are written by someone, they are not magically built in Julia. So you can closely follow the examples written here. Generally, you will need to first create a metric type
```
struct MyMetric <: Metric end
```
You will then need to define an instance of `evaluate`:
```
function evaluate(dist::MyMetric,a::AbstractArray,b::AbstractArray)
    # how to compute dist(a,b)
end
```
You then have automatically generic `pairwise` and `colwise` functions, which you can specifically optimize if needed, analogously to `pairwise(::Euclidean,A,B)` etc. It is probably easiest if you start with the whole block of code related to one specific metric that is somewhat close to your required structure, i.e., including convenience functions like, e.g., `euclidean(a,b)`, and modify it according to your own needs. Hope this helps.

000Justin000:
Hello Daniel,

Thank you kindly for your reply.

Actually what I am asking is that whether an additional metric can be added without touching the original files of the package, as I want to later distribute my package as well. Is there an interface to do that?

Thanks,
Junteng



Sent from my iPhone

On Jun 20, 2018, at 12:36 PM, Daniel Karrasch <notifications@github.com<mailto:notifications@github.com>> wrote:


Sure, all metrics in the package are written by someone, they are not magically built in Julia. So you can closely follow the examples written here. Generally, you will need to first create a metric type

struct MyMetric <: Metric end


You will then need to define an instance of evaluate:

function evaluate(dist::MyMetric,a::AbstractArray,b::AbstractArray)
    # how to compute dist(a,b)
end


You then have automatically generic pairwise and colwise functions, which you can specifically optimize if needed, analogously to pairwise(::Euclidean,A,B) etc. It is probably easiest if you start with the whole block of code related to one specific metric that is somewhat close to your required structure, i.e., including convenience functions like, e.g., euclidean(a,b), and modify it according to your own needs. Hope this helps.

dkarrasch:
[...]

PR #150, opened 2019-10-03
Extend parameter trait to wmetrics

dkarrasch:
This is conceptually nothing but a rebase of #135, so (almost) all the credit goes to @devmotion.

This first commit simply moves wmetrics.jl to metrics.jl, the second rearranges a few docstrings. Functional changes start only in commit 3.

EDIT: Closes #135, but we should have @devmotion as a co-author.


dkarrasch:
I think this is ready for review. As for the little regression benchmark from #107, this is what I get on this branch:
```julia
julia> @btime colwise(dd, Xcol,Ycol);
  798.898 μs (2 allocations: 781.33 KiB)

julia> @btime colwise_naive(Xcol,Ycol);
  834.077 μs (2 allocations: 781.33 KiB)
```
The little gap is consistent and in the "desired direction". What #135 was also doing and what is included here is fixing some length checking and return type issues for the weighted metrics. Other than that this establishes the parameter trait across all `UnionMetrics` (including the weighted ones) and adds a few scalar methods.

src/metrics.jl
```diff
@@ -5,24 +5,107 @@
 #   Metric types
 #
 ###########################################################
+const RealAbstractArray{T <: Real} =  AbstractArray{T}
```
nalimilan: Isn't `AbstractArray{<:Real}` short enough?

dkarrasch: Yes, this was used in the former `wmetrics.jl`, but is unnecessary.

dkarrasch:
It would be great if we could merge these changes this time before other dramatic PRs come in. Therefore, here's a gentle "bump".

KristofferC:
Correct me if I am wrong but by doing stuff like:

```
const weightedmetrics = (WeightedEuclidean,WeightedSqEuclidean,WeightedCityblock,WeightedMinkowski,WeightedHamming)
const UnionWeightedMetrics = Union{weightedmetrics...}
const UnionMetrics = Union{metrics...,weightedmetrics...}
```

and then dispatching on `UnionMetrics` means it is impossible for anyone else to add their own `WeightedMetrics`.
What is the reason we cannot use standard multiple dispatch instead of these hardcoded list of types?

src/metrics.jl
```diff
@@ -348,31 +397,31 @@ end
...
 
-# Correlation Dist
+# CorrDist
 _centralize(x::AbstractArray) = x .- mean(x)
 (dist::CorrDist)(a::AbstractArray, b::AbstractArray) = CosineDist()(_centralize(a), _centralize(b))
+(dist::CorrDist)(a::Number, b::Number) = CosineDist()(zero(mean(a)), zero(mean(b)))
 corr_dist(a::AbstractArray, b::AbstractArray) = CorrDist()(a, b)
-result_type(::CorrDist, Ta::Type, Tb::Type) = result_type(CosineDist(), Ta, Tb)
+corr_dist(a::Number, b::Number) = CorrDist()(a, b)
+# result_type(::CorrDist, Ta::Type, Tb::Type) = result_type(CosineDist(), Ta, Tb)
```
nalimilan: ?

src/metrics.jl
```diff
@@ -100,216 +182,174 @@ struct MeanSqDeviation <: SemiMetric end
...
+                    s = eval_reduce(d, s, eval_op(d, ai, bi))
+                end
             end
         end
         return eval_end(d, s)
     end
 end
 
-@inline function _evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray)
+@inline Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray, p::AbstractArray)
```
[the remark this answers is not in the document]
dkarrasch: Pretty much every use of `Base.@propagate_inbounds` is decorated by `@inline`, so I'll remove all of those `@inline`s.

src/metrics.jl
```diff
@@ -100,216 +182,174 @@ struct MeanSqDeviation <: SemiMetric end
...
                 s = eval_reduce(d, s, eval_op(d, ai, bi))
             end
         else
-            @simd for I in 1:length(a)
-                aI = a[I]
-                bI = b[I]
-                pI = p[I]
-                s = eval_reduce(d, s, eval_op(d, aI, bI, pI))
+            if size(a) == size(b)
+                @simd for I in eachindex(a, b)
```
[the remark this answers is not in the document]
dkarrasch: The `@inbounds` appears a few lines up. I'll merge the two "branches" as you proposed.

src/metrics.jl
dkarrasch: Sorry, I forgot to delete it. This is simply unnecessary, because the generic code already does the job.

KristofferC:
Yeah, I'm sorry if my comment wasn't directly related to the PR but more of a comment about the current state of affairs.

There is a lot of code movement here and it is a bit unclear to me what is the core change in this PR? Is it just a bug fix and the added tests fail on master?

dkarrasch:
First of all, this PR joins the weighted metrics with the `UnionMetrics`. To achieve this, `parameters` is generalized towards a "trait" for all union metrics: the default being `nothing`, and the `d.periods` for the periodic Euclidean distance, and `d.weights` for the classic weighted distances. By doing so, a few minor bugs/inconsistencies in the weighted metrics are fixed: https://github.com/JuliaStats/Distances.jl/pull/135#issuecomment-496818142 Finally, this adds a few `Number` methods for some metrics that were previously missing, and some corresponding tests. A few other tests are simply strengthened to also test for the exact return type. Other than that, a few docstrings are moved for more consistency.

Now that I understand that your comment was not related to the actual PR, I think I understand what you mean, and I was thinking about it when I was writing the periodic distance. Should I introduce some abstract types like `UnionMetric <: Metric` and `UnionSemiMetric <: SemiMetric`, make the distances subtypes of the corresponding `UnionXYZ` and dispatch on that? Or shall we have that in a separate PR?

KristofferC:
> Should I introduce some abstract types like UnionMetric <: Metric and UnionSemiMetric <: SemiMetric

Yes, that makes sense to me.

> Or shall we have that in a separate PR?

I think that is the best.

src/metrics.jl
```diff
@@ -100,216 +182,156 @@ struct MeanSqDeviation <: SemiMetric end
...
-        length(a) != length(p) && throw(DimensionMismatch("arrays have length $(length(a)) but parameters have length $(length(p))."))
-    end
     if length(a) == 0
         return zero(result_type(d, a, b))
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if p === nothing
+        if IndexStyle(a, b) === IndexLinear() || size(a) == size(b)
             @simd for I in 1:length(a)
```
nalimilan: This is more efficient for `IndexCartesian`:
```suggestion
            @simd for I in eachindex(a, b)
```

dkarrasch: Indeed, in the little benchmark it's performing faster. 👍

dkarrasch: Wow, that is actually very interesting, because the reason `1:length(a)` (to index into _both_ `a` _and_ `b` was added in the first place was https://github.com/JuliaLang/julia/issues/28226. However, now `eachindex(a, b)` seems to work nicely:
```julia
julia> @btime colwise(dd, Xcol,Ycol);
  923.614 μs (2 allocations: 781.33 KiB)

julia> @btime colwise_naive(Xcol,Ycol);
  975.183 μs (2 allocations: 781.33 KiB)
```

dkarrasch:
Bump. 😇

src/metrics.jl
```diff
@@ -222,7 +222,7 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     @inbounds begin
         s = eval_start(d, a, b)
         if IndexStyle(a, b) === IndexLinear() || size(a) == size(b)
-            @simd for I in 1:length(a)
+            @simd for I in eachindex(a, b)
```
KristofferC: Note https://github.com/JuliaLang/julia/issues/28226#issuecomment-406977072.

Is that no longer the case?

dkarrasch: I tested https://github.com/JuliaStats/Distances.jl/pull/150#issuecomment-538310651, and obtained the results above. Wasn't this the test that would determine the slowdown?

dkarrasch: Sorry, I meant https://github.com/JuliaStats/Distances.jl/pull/107#issuecomment-473232505

dkarrasch: I get that `colwise` is consistently and measurably faster than `naivecolwise` in that test.

KristofferC: I guess it is fixed then...

dkarrasch: I'll quickly double-check on Julia v1.0...

dkarrasch: Alright, I have tested my but off, and I can't seem to be able to reproduce the slowdown anymore, even on Julia v1.0. While the issue https://github.com/JuliaLang/julia/issues/28226 persists even on nightlies, the issue https://github.com/JuliaStats/Distances.jl/issues/100 does not re-occur with this PR, neither on Julia v1.0.5 nor v1.3 or nightlies. AFAICT, we wouldn't introduce a slowdown with the change in the line commented on above.

dkarrasch:
Gentle bump.

dkarrasch:
Bump again. `colwise` is getting ever more faster on each new Julia version. In this test https://github.com/JuliaStats/Distances.jl/pull/107#issuecomment-473232505 this branch is significantly faster than `colwise_naive`!

I'd like to continue with the transition to `Union[Pre/Semi]Metric` types, which should help resolve #126 and #95.

devmotion:
Bumping again :slightly_smiling_face: What is holding this PR back?

dkarrasch:
Gentle bump once more.

dkarrasch:
Please, could we make some progress here? There are new PRs starting to pile up (like #161) that would make it unnecessarily tedious to merge this after those.

PR #158, opened 2020-02-16
Add p-Wasserstein distance

### LINKED ISSUE #37: Add Earth Mover's Distance/Wasserstein metric
EMD is useful for calculating distances between histograms or distributions. It would be cool to see it added to Distances.jl

https://en.wikipedia.org/wiki/Earth_mover's_distance

Sh4pe:
This PR adds the p-Wasserstein distance (a.k.a. Earth Mover Distance). Calculating the p-Wasserstein entails solving a linear program. This is done using JuMP.jl with the Cbc.jl optimizer.

Fixes #37

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,6 +3,8 @@ uuid = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 version = "0.8.2"
 
 [deps]
+Cbc = "9961bab8-2fa3-5c5a-9d89-47fab24efd76"
+JuMP = "4076af6c-e467-56ae-b986-b466b2749572"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 
--- a/README.md
+++ b/README.md
@@ -38,6 +38,7 @@ This package also provides optimized functions to compute column-wise and pairwi
 * Normalized root mean squared deviation
 * Bray-Curtis dissimilarity
 * Bregman divergence
+* Wasserstein distance
 
 For `Euclidean distance`, `Squared Euclidean distance`, `Cityblock distance`, `Minkowski distance`, and `Hamming distance`, a weighted version is also provided.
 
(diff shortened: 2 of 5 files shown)


johnnychen94:
I would say the main issue of introducing this PR is its huge dependency. You know, at present this package is quite small

```julia
julia> @time using Distances
  0.078969 seconds (52.74 k allocations: 3.509 MiB)
```

Sh4pe:
> I would say the main issue of introducing this PR is its huge dependency. You know, at present this package is quite small
> 
> ```julia
> julia> @time using Distances
>   0.078969 seconds (52.74 k allocations: 3.509 MiB)
> ```
> 
> Probably this package isn't the best place to put the codes to

Yes, this PR introduces a lot of dependencies. But since you need to solve a (possibly large) linear program in order to calculate the Wasserstein distance, you need to introduce at least _some_ dependencies (granted that you don't want to re-implement a linear program solver on your own)...

So what do you think would be a better place for the code? 

And what do others think?

Sh4pe:
Another implementation (though solved quite differently) is in #159. Don't really know about the implementation there though.

devmotion:
FYI: https://github.com/JuliaOptimalTransport/OptimalTransport.jl contains specialized implementations for the 1D solution, both for discrete and continuous distributions, but also a pure Julia implementation for solving the optimal transport problem with discrete measures in arbitrary dimensions (and soon there will be also a specialized implementation for multivariate normal distributions). We just added [`wasserstein`](https://juliaoptimaltransport.github.io/OptimalTransport.jl/dev/#OptimalTransport.wasserstein) as an interface for computing the p-Wasserstein distance.

nalimilan:
Adding JuMP as a dependency to Distances.jl would certainly be a problem. As noted at https://github.com/JuliaStats/Distances.jl/pull/159#issuecomment-649856122, the 1D Wasserstein distance has a closed form solution, and Distances.jl only includes 1D distances, so we should use that.

devmotion:
We started splitting OptimalTransport.jl into smaller subpackages, the Wasserstein distances, based on MOI in the general case and including special cases as 1D and e.g. Gaussian distributions, are moved to https://github.com/JuliaOptimalTransport/ExactOptimalTransport.jl.

PR #161, opened 2020-03-30
Ensure non-negativity of pairwise computations

devmotion:
This PR supersedes https://github.com/JuliaStats/Distances.jl/pull/13 and fixes the problem that currently pairwise evaluations of the (weighted) squared Euclidean distance sometimes yield negative values:
```julia
julia> using Distances

julia> a = [0.3 0.3 + eps()]
1×2 Array{Float64,2}:
 0.3  0.3

julia> pairwise(SqEuclidean(), a; dims = 2)
2×2 Array{Float64,2}:
  0.0          -2.77556e-17
 -2.77556e-17   0.0

julia> pairwise(WeightedSqEuclidean([1.0]), a; dims = 2)
2×2 Array{Float64,2}:
  0.0          -2.77556e-17
 -2.77556e-17   0.0
```
The fix is the same that is already applied for `CosineDist` at the moment.

Additionally, I made the computation of the pairwise computation of the Euclidean distance more similar to the computation of the squared Euclidean distance by adding an additional branch with `@simd` annotation. I assume the problems with negative results have not been observed for the Euclidean distance due to the missing `@simd` annotation.


dkarrasch:
Does the `@simd` branch speed up certain computations?

devmotion:
It's supposed to since it signals the compiler that it is OK to perform additional optimizations, but I haven't benchmarked it.

As I wrote above, this PR does not contain any sophisticated new implementation - I just made the implementation of Euclidean consistent with the existing implementation of SqEuclidean, and applied the fix for non-negativity to (Weighted)SqEuclidean and Euclidean that is currently used by CosineDist.

BTW it seems the test failure is not related to the changes in this PR.

dkarrasch:
I've tested
```julia
using Distances, BenchmarkTools
A = rand(3, 10000);
@btime pairwise($(Euclidean()), $A; dims=2);
@btime pairwise($(SqEuclidean()), $A; dims=2);
```
and don't notice any consistent change pattern in runtime, certainly no consistent slowdown, but also no speedup. Shall we merge this?

devmotion:
Bump

devmotion:
It seems this fix might also help to avoid a hack in Zygote https://github.com/FluxML/Zygote.jl/pull/609.

PR #163, opened 2020-04-10
turn UnionMetrics into types

dkarrasch:
As discussed in https://github.com/JuliaStats/Distances.jl/pull/150#issuecomment-541606075, this allows for dispatching on `Union*Metric` types, and should facilitate hooking into the `eval_***` and `parameters` business.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -2,11 +2,23 @@
 
 ###########################################################
 #
+#   Abstract metric types
+#
+###########################################################
+
+abstract type UnionPreMetric <: PreMetric end
+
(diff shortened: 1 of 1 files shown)


KristofferC:
Makes sense to me, I wonder how breaking this is.

dkarrasch:
I don't think it's breaking. Previously, the `UnionMetrics` and `evaluate(::UnionMetrics, a, b)` framework were closed for external packages. So this change here is, as far as I see, a truly new feature. Maybe I should add some documentation on how to write a new `UnionMetric`?

KristofferC:
Could be added later

PR #164, opened 2020-04-13
Enhancements: allow OffsetArrays and integer arrays in CosineDist

### LINKED ISSUE #93: compatibility with OffsetArrays
Currently not compatible with OffsetArrays, because of calls to `size` (possibly among other things):

```
julia> using Distances, OffsetArrays

julia> x,y = rand(10), rand(10)
([0.122568, 0.558197, 0.890631, 0.622067, 0.549109, 0.285662, 0.143514, 0.987727, 0.971258, 0.287773], [0.429365, 0.433621, 0.287694, 0.452117, 0.872341, 0.442937, 0.0165311, 0.74837, 0.648707, 0.0330926])

julia> corr_dist(x,y)
0.44865252243622267

julia> xoff = OffsetVector(x, -5:4)
OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}} with indices -5:4:
 0.122568
 0.558197
 0.890631
 0.622067
 0.549109
 0.285662
 0.143514
 0.987727
 0.971258
 0.287773

julia> corr_dist(xoff, y)
ERROR: size not supported for arrays with axes (-5:4,); see http://docs.julialang.org/en/latest/devdocs/offset-arrays/
Stacktrace:
 [1] errmsg(::OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}}) at /home/dave/.julia/v0.6/OffsetArrays/src/OffsetArrays.jl:91
 [2] evaluate at /home/dave/.julia/v0.6/Distances/src/metrics.jl:170 [inlined]
 [3] cosine_dist at /home/dave/.julia/v0.6/Distances/src/metrics.jl:257 [inlined]
 [4] evaluate at /home/dave/.julia/v0.6/Distances/src/metrics.jl:261 [inlined]
 [5] corr_dist(::OffsetArrays.OffsetArray{Float64,1,Array{Float64,1}}, ::Array{Float64,1}) at /home/dave/.julia/v0.6/Distances/src/metrics.jl:264
```

This could probably be fixed pretty easily by the suggestion from the OffsetArrays readme to use internal helper functions `_size(A::AbstractArray) = map(length, indices(A))`.

### LINKED ISSUE #106: pairwise CosineDistance with Int matrices as argument throws error
Here's a quick example:

```julia
using Distances
A = rand(1:100, (100, 100));
pairwise(CosineDist(), A)

ERROR: InexactError: Int64(Int64, 595.1344385934997)
Stacktrace:
 [1] Type at ./float.jl:700 [inlined]
 [2] convert at ./number.jl:7 [inlined]
 [3] setindex! at ./array.jl:769 [inlined]
 [4] macro expansion at /Users/isaac/.julia/packages/Distances/nLAdT/src/common.jl:96 [inlined]
 [5] macro expansion at ./simdloop.jl:73 [inlined]
 [6] sqrt!(::Array{Int64,1}) at /Users/isaac/.julia/packages/Distances/nLAdT/src/common.jl:95
 [7] pairwise!(::Array{Float64,2}, ::CosineDist, ::Array{Int64,2}) at /Users/isaac/.julia/packages/Distances/nLAdT/src/metrics.jl:596
 [8] pairwise(::CosineDist, ::Array{Int64,2}) at /Users/isaac/.julia/packages/Distances/nLAdT/src/generic.jl:127
 [9] top-level scope at none:0
```

It looks like this is happening because of [this line](https://github.com/JuliaStats/Distances.jl/blob/0167683d5fddf7814e473ec4a6b8c07a6956fccc/src/metrics.jl#L596), since `sumsq_percol` returns return a integer vector for an integer matrix. I did a little benchmarking to see the costs of using `sqrt.` instead of `sqrt!` version, and don't see much difference.

dkarrasch:
Closes #106. Closes #93.

The logic in the `UnionMetric`'s `_evaluate` wasn't able to distinguish the case when one array had offset indices. But I'm not sure if the logic here is too strict. I don't see any slowdown, though benchmarking is a bit nontrivial. I've used this for benchmarking:
```julia
using Distances, OffsetArrays, BenchmarkTools

for m in (2, 10, 30)
  x, y, p = rand(m), rand(m), rand(m)
  d = Euclidean()
  dp = PeriodicEuclidean(p)
  @btime $d($x, $y);
  @btime $dp($x, $y);
end
```
Timings vary a bit, but I don't see a clear pattern of regression. Also, the infamous benchmark from #107 is clearly won by `colwise`.

Moreover
```julia
julia> using Distances, OffsetArrays

julia> x, y = rand(10), rand(10)
([0.02801362061758983, 0.8939414820779741, 0.711232152659703, 0.9910084998815372, 0.09319994918278351, 0.4034886165223559, 0.22832132159907648, 0.6169286534167804, 0.9309683705377529, 0.6998361254000887], [0.2479258273830649, 0.2573826000306132, 0.765649154420025, 0.06495363114307429, 0.41802555664920504, 0.8972858186710049, 0.4236709655992299, 0.3596080783013793, 0.4169463881238391, 0.7490678314251886])

julia> corr_dist(x,y)
1.1160543248280592

julia> xoff = OffsetVector(x, -5:4)
10-element OffsetArray(::Array{Float64,1}, -5:4) with eltype Float64 with indices -5:4:
 0.02801362061758983
 0.8939414820779741
 0.711232152659703
 0.9910084998815372
 0.09319994918278351
 0.4034886165223559
 0.22832132159907648
 0.6169286534167804
 0.9309683705377529
 0.6998361254000887

julia> corr_dist(xoff, y)
1.1160543248280592
```
and
```julia
julia> using Distances

julia> A = rand(1:100, (100, 100));

julia> pairwise(CosineDist(), A; dims=2)
100×100 Array{Float64,2}:
...
```

### DIFF
--- a/src/common.jl
+++ b/src/common.jl
@@ -91,9 +91,6 @@ end
 #
 ###########################################################
 
-_sqrt(a::AbstractArray{<:Integer}) = sqrt.(a)
-_sqrt(a) = sqrt!(a)
-
 function sqrt!(a::AbstractArray)
     @simd for i in eachindex(a)
         @inbounds a[i] = sqrt(a[i])
@@ -102,12 +99,22 @@ function sqrt!(a::AbstractArray)
 end
 
 function sumsq_percol(a::AbstractMatrix{T}) where {T}
-    m = size(a, 1)
     n = size(a, 2)
     r = Vector{T}(undef, n)
     for j = 1:n
         aj = view(a, :, j)
         r[j] = dot(aj, aj)
+    end
+    return r
+end
+
+function norm_percol(a::AbstractMatrix{T}) where {T}
+    n = size(a, 2)
+    √T = typeof(sqrt(oneunit(T)))
+    r = Vector{√T}(undef, n)
+    for j in 1:n
+        aj = view(a, :, j)
+        r[j] = sqrt(dot(aj, aj))
     end
     return r
 end
--- a/src/common.jl
+++ b/src/common.jl
@@ -101,7 +101,7 @@ end
 function sumsq_percol(a::AbstractMatrix{T}) where {T}
     n = size(a, 2)
     r = Vector{T}(undef, n)
-    for j = 1:n
+    @simd for j in 1:n
         aj = view(a, :, j)
         r[j] = dot(aj, aj)
     end
@@ -112,7 +112,7 @@ function norm_percol(a::AbstractMatrix{T}) where {T}
     n = size(a, 2)
     √T = typeof(sqrt(oneunit(T)))
     r = Vector{√T}(undef, n)
-    for j in 1:n
+    @simd for j in 1:n
         aj = view(a, :, j)
         r[j] = sqrt(dot(aj, aj))
     end
--- a/src/common.jl
+++ b/src/common.jl
@@ -91,6 +91,9 @@ end
 #
 ###########################################################
 
+_sqrt(a::AbstractArray{<:Integer}) = sqrt.(a)
+_sqrt(a) = sqrt!(a)
+
 function sqrt!(a::AbstractArray)
     @simd for i in eachindex(a)
         @inbounds a[i] = sqrt(a[i])
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -236,7 +236,7 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if IndexStyle(a, b) === IndexLinear() && eachindex(a) == eachindex(b)
+        if IndexStyle(a, b) === IndexLinear() && axes(a) == axes(b)
             @simd for I in eachindex(a, b)
                 ai = a[I]
                 bi = b[I]
@@ -265,7 +265,7 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if IndexStyle(a, b, p) === IndexLinear() && eachindex(a) == eachindex(b) == eachindex(p)
+        if IndexStyle(a, b, p) === IndexLinear() && axes(a) == axes(b) == axes(p)
             @simd for I in eachindex(a, b, p)
                 ai = a[I]
                 bi = b[I]
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -236,7 +236,7 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if IndexStyle(a, b) === IndexLinear() && axes(a) == axes(b)
+        if (IndexStyle(a, b) === IndexLinear() && eachindex(a) == eachindex(b)) || axes(a) == axes(b)
             @simd for I in eachindex(a, b)
                 ai = a[I]
                 bi = b[I]
@@ -265,7 +265,8 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if IndexStyle(a, b, p) === IndexLinear() && axes(a) == axes(b) == axes(p)
+        if (IndexStyle(a, b, p) === IndexLinear() && eachindex(a) == eachindex(b) == eachindex(p)) ||
+                axes(a) == axes(b) == axes(p)
             @simd for I in eachindex(a, b, p)
                 ai = a[I]
                 bi = b[I]
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -786,8 +786,8 @@ function _pairwise!(r::AbstractMatrix, dist::CosineDist,
                     a::AbstractMatrix, b::AbstractMatrix)
     m, na, nb = get_pairwise_dims(r, a, b)
     mul!(r, a', b)
-    ra = _sqrt(sumsq_percol(a))
-    rb = _sqrt(sumsq_percol(b))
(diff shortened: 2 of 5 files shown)


src/metrics.jl
```diff
@@ -236,7 +236,7 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if IndexStyle(a, b) === IndexLinear() || size(a) == size(b)
+        if IndexStyle(a, b) === IndexLinear() && eachindex(a) == eachindex(b)
```
dkarrasch, line 239: This is the crucial part: it makes sure that `eachindex(a, b)` in the next line is constructible.

src/common.jl
```diff
@@ -91,6 +91,9 @@ end
 #
 ###########################################################
 
+_sqrt(a::AbstractArray{<:Integer}) = sqrt.(a)
```
dkarrasch, line 94: This one is only used for intermediate results, not for in-place return values.

dkarrasch:
Does anyone have nontrivial (i.e., cases where `IndexStyle(a, b) === IndexLinear() && eachindex(a) == eachindex(b)` fails but you still expect the `@simd` branch to work) test cases to make sure that the new branching logic doesn't slow down anything? I'd be happy to do the benchmarking to come to some conclusion here.

src/common.jl
devmotion: Maybe it would be useful to change `sumsq_percol` instead, so that the resulting array contains floating point numbers if one intends to compute the square root of these numbers later on? Then one could avoid to allocate these additional arrays here.

dkarrasch:
I think we are in the position to make a v0.9 release. I wonder if we should include this PR also, or whether we should keep it for review and make the release right away. Opinions?

src/common.jl
```diff
@@ -109,6 +108,17 @@ function sumsq_percol(a::AbstractMatrix{T}) where {T}
...
+function norm_percol(a::AbstractMatrix{T}) where {T}
+    n = size(a, 2)
+    √T = typeof(sqrt(oneunit(T)))
+    r = Vector{√T}(undef, n)
+    for j in 1:n
+        aj = view(a, :, j)
+        r[j] = sqrt(dot(aj, aj))
+    end
+    return r
+end
```
johnnychen94, line 120: I find this version faster than the for-loop version

```julia
function norm_percol_(a)
    map(eachcol(a)) do aj
        sqrt(dot(aj, aj))
    end
end

julia> a = rand(512, 512);

julia> @btime norm_percol($a);
  71.526 μs (1538 allocations: 44.14 KiB)

julia> @btime norm_percol_($a);
  49.120 μs (516 allocations: 28.19 KiB)
```

src/metrics.jl
johnnychen94: Looks like it makes no difference, but I guess you actually want to say `axes(a) == axes(b)`?

src/common.jl
```diff
@@ -109,6 +108,17 @@ function sumsq_percol(a::AbstractMatrix{T}) where {T}
     return r
 end
 
+function norm_percol(a::AbstractMatrix{T}) where {T}
+    n = size(a, 2)
+    √T = typeof(sqrt(oneunit(T)))
+    r = Vector{√T}(undef, n)
+    for j in 1:n
```
johnnychen94, line 115: I didn't get the need of `norm_percol` here given that it's equivalent to `sqrt!(sumsq_percol(a))`

FYI

```diff
-    for j in 1:n
+    @simd for j in 1:n
```

This makes the op 2x faster in my machine. I think we could do the same to `sumsq_percol` as well.

dkarrasch: When `a` is a matrix of `Int`s, then `sumsq_percol(a)` is a vector of `Int`s, and then you can't compute the `sqrt` in-place, see #106. So, `norm_percol` allocates a vector of the required type to store the square root. I'll add the `@simd` macro.

src/metrics.jl
dkarrasch: You mean replacing the entire condition or just the second part?

src/metrics.jl
```diff
@@ -236,7 +236,7 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
     @inbounds begin
         s = eval_start(d, a, b)
-        if IndexStyle(a, b) === IndexLinear() || size(a) == size(b)
+        if IndexStyle(a, b) === IndexLinear() && axes(a) == axes(b)
```
johnnychen94, line 239: I believe it should still be `||` here.

An example of a non-linear indexing array is `PaddedView`:

```julia
x = rand(2, 3);
px = PaddedView(-1, x, (4, 4));
IndexStyle(px); # IndexCartesian()
axes(px); # (Base.OneTo(4), Base.OneTo(4))

y = rand(3, 2);
py = PaddedView(-1, y, (4, 4));
axes(py) # (Base.OneTo(4), Base.OneTo(4))

axes(py) == axes(px) # true
```

In this case, they could share the same index `I` in the loop and hence we don't need to use the fallback implementation, where axes of arrays are different.

dkarrasch: If I put `||`, then the `OffsetArray` fix won't work, because the example in #93 has linearly indexed vectors. I'll go with
```julia
if (IndexStyle(a, b) === IndexLinear() && eachindex(a) == eachindex(b)) || axes(a) == axes(b)
```
This would catch the different indices with OffsetArrays, but allow cartesian-indexing as long as the axes are the same.

dkarrasch: With this PR, you can do
```julia
Euclidean()(px, py)
```
and it works. According to the latest logic, it uses the `@simd` branch.

dkarrasch:
Alright, tests (that fail on master) are added and pass. If there are no objections, I'd merge this and release v0.9 soonish.

PR #166, opened 2020-06-18
Fixed grammar

pauljurczak:


README.md
```diff
@@ -55,7 +55,7 @@ r = evaluate(dist, x, y)
 r = dist(x, y)
 ```
 
-Here, `dist` is an instance of a distance type. For example, the type for Euclidean distance is `Euclidean` (more distance types will be introduced in the next section), then you can compute the Euclidean distance between `x` and `y` as
+Here, `dist` is an instance of a distance type. For example, if the type for Euclidean distance is `Euclidean` (more distance types will be introduced in the next section), then you can compute the Euclidean distance between `x` and `y` as
```
nalimilan, line 58: I don't think "if" is right here, as this is just a statement. Maybe this?
```suggestion
Here, `dist` is an instance of a distance type: for example, the type for Euclidean distance is `Euclidean` (more distance types will be introduced in the next section). You can compute the Euclidean distance between `x` and `y` as
```

pauljurczak: You are right. Your version seems to be the original intent.

PR #176, opened 2020-09-01
add default value of Earth's radius to Haversine

mkborregaard:
I don't think it should be necessary to look up the radius of Earth every time one wants to use the Haversine formula, so I've added the equatorial radius of Earth in meters as a default. I hope most imperial unit users will agree that metric is a reasonable default for scientific libraries. Of course Earth has different radii depending on where it is measured, as it isn't a perfect sphere. Uses sensitive to this should use more specific values. FYI this same default value is used in R. source https://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -1,10 +1,12 @@
 """
-    Haversine(radius)
+    Haversine([radius])
 
 The haversine distance between two locations on a sphere of given `radius`.
 
 Locations are described with longitude and latitude in degrees.
 The computed distance has the same units as that of the radius.
+The default value is 6371000 meters, which is the mean volumetric
+radius of Earth (source https://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html).
 """
 struct Haversine{T<:Real} <: Metric
     radius::T
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -9,6 +9,7 @@ The computed distance has the same units as that of the radius.
 struct Haversine{T<:Real} <: Metric
     radius::T
 end
+Haversine() = Haversine(6378137.0)
 
 function (dist::Haversine)(x, y)
     length(x) == length(y) == 2 || haversine_error(dist)
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -9,7 +9,7 @@ The computed distance has the same units as that of the radius.
 struct Haversine{T<:Real} <: Metric
     radius::T
 end
-Haversine() = Haversine(6378137.0)
+Haversine() = Haversine(6371000.0)
 
 function (dist::Haversine)(x, y)
     length(x) == length(y) == 2 || haversine_error(dist)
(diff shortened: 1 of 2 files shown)


mkborregaard:
An additional concern could be that the function assumes the first point to be longitude, the second latitude, but that's not documented anywhere. Add to readme?

dkarrasch:
> An additional concern could be that the function assumes the first point to be longitude, the second latitude, but that's not documented anywhere. Add to readme?

I think it's mentioned in the docstring? But adding a short remark to the README may be helpful as well. Also, the default value including its unit should be stated in the docstring.

src/haversine.jl
```diff
@@ -33,6 +34,6 @@ function (dist::Haversine)(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple)
     2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
 end
 
-haversine(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple, radius::Real) = Haversine(radius)(x, y)
+haversine(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple, radius::Real = 6378137.0) = Haversine(radius)(x, y)
```
nalimilan, line 37: Convention in this package seems to be to use no spaces:
```suggestion
haversine(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple, radius::Real=6378137.0) =
    Haversine(radius)(x, y)
```

src/haversine.jl
```diff
@@ -9,6 +9,7 @@ The computed distance has the same units as that of the radius.
 struct Haversine{T<:Real} <: Metric
     radius::T
 end
+Haversine() = Haversine(6378137.0)
```
nalimilan, line 12: Maybe it would be better to make this a `Float32` or an `Int32` to avoid promoting `Float32` values to `Float64` unnecessarily?

src/haversine.jl
```diff
@@ -9,6 +9,7 @@ The computed distance has the same units as that of the radius.
 struct Haversine{T<:Real} <: Metric
```
nalimilan, line 9: The docstring above should mention the default, say what it corresponds to exactly, and give links to references (e.g. Wikipedia).

nalimilan:
Good idea. Though most references I found use 6371 km, which is the [mean radius](https://en.wikipedia.org/wiki/Earth_radius#Mean_radius) and recommended [according to Rosetta code](https://rosettacode.org/wiki/Haversine_formula):
https://www.movable-type.co.uk/scripts/latlong.html
https://community.esri.com/t5/coordinate-reference-systems/distance-on-a-sphere-the-haversine-formula/ba-p/902128
https://www.igismap.com/haversine-formula-calculate-geographic-distance-earth/
https://www.geeksforgeeks.org/haversine-formula-to-find-distance-between-two-points-on-a-sphere/
https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.haversine_distances.html

References that mention 6378 km often also mention 6371 km and indicate it depends on where you are on earth. So it seems 6371 km may be more standard? See also https://github.com/mapbox/turf-swift/issues/26 and linked issues for a discussion.

dkarrasch:
I didn't look too closely, but I agree that, in my area, people seem to use 6371 km, which, according to the Earth Fact Sheet linked in the OP, corresponds to the volumetric mean radius.

src/haversine.jl
mkborregaard: done

mkborregaard:
Sorry for delay - I've edited as requested

nalimilan:
Sorry for not commenting earlier, but this needs a test.

dkarrasch:
I have tests in #203.

PR #179, opened 2020-09-29
simplify result_type on UnionMetrics

johnnychen94:
#150 has unfortunately broken the glue code in ImageDistances because it specializes the `result_type` for `UnionMetrics` without defining the method for `Type`.

https://github.com/JuliaStats/Distances.jl/blob/44036b573ec85287022f4368c4e1e279698bd031/src/generic.jl#L35

This PR adds these two methods:

```julia
result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, ::Nothing)
result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, p)
```

and https://github.com/JuliaImages/ImageDistances.jl/pull/46 gets fixed by this.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -213,12 +213,12 @@ for dist in weightedmetrics
     @eval parameters(d::$dist) = d.weights
 end
 
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray) =
-    result_type(dist, a, b, parameters(dist))
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, ::Nothing) =
-    typeof(_evaluate(dist, oneunit(eltype(a)), oneunit(eltype(b))))
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, p) =
-    typeof(_evaluate(dist, oneunit(eltype(a)), oneunit(eltype(b)), oneunit(eltype(p))))
+result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}) where {Ta,Tb} =
+    result_type(dist, Ta, Tb, parameters(dist))
+result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, ::Nothing) where {Ta,Tb} =
+    typeof(_evaluate(dist, oneunit(Ta), oneunit(Tb)))
+result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, p) where {Ta,Tb} =
+    typeof(_evaluate(dist, oneunit(Ta), oneunit(Tb), oneunit(eltype(p))))
 
 Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray)
     _evaluate(d, a, b, parameters(d))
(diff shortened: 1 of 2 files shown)


src/metrics.jl
```diff
@@ -213,12 +213,12 @@ for dist in weightedmetrics
...
 end
 
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray) =
-    result_type(dist, a, b, parameters(dist))
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, ::Nothing) =
-    typeof(_evaluate(dist, oneunit(eltype(a)), oneunit(eltype(b))))
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, p) =
-    typeof(_evaluate(dist, oneunit(eltype(a)), oneunit(eltype(b)), oneunit(eltype(p))))
+result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, p=nothing) =
+    result_type(dist, eltype(a), eltype(b), isnothing(p) ? parameters(dist) : p)
```
devmotion, line 217: It seems there is no need for `isnothing` here if you just define
```suggestion
result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, p=parameters(dist)) =
    result_type(dist, eltype(a), eltype(b), p)
```
BTW depending on your use case in ImageDistances (which is not completely clear to me) it might be sufficient to define
```suggestion
result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}) where {Ta,Tb} =
    result_type(dist, Ta, Tb, parameters(dist))
```
since the fallback is `result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))`.

dkarrasch: Indeed, the second suggestion works just fine (without the compat stuff), after you remove the `p` from the arguments here:

https://github.com/JuliaStats/Distances.jl/blob/d547d3a78db98411a902fe5382243ece00bfdcc3/src/metrics.jl#L264

dkarrasch:
@johnnychen94 Would you be able to finish this soon? Together with #182 this would make up a nice v0.9.2 release that you can then build on.

johnnychen94:
Thanks for the reminder, yes I'll make it tomorrow.

The suggestion seems reasonable a change but I'll need some time to check it.

src/metrics.jl
```diff
@@ -213,12 +213,14 @@ for dist in weightedmetrics
...
 end
 
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray) =
-    result_type(dist, a, b, parameters(dist))
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, ::Nothing) =
-    typeof(_evaluate(dist, oneunit(eltype(a)), oneunit(eltype(b))))
-result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, p) =
-    typeof(_evaluate(dist, oneunit(eltype(a)), oneunit(eltype(b)), oneunit(eltype(p))))
+result_type(dist::UnionMetrics, a::AbstractArray, b::AbstractArray, p=parameters(dist)) =
+    result_type(dist, eltype(a), eltype(b), p)
```
dkarrasch, line 217: For clarity, I'd suggest to remove this method. Let `result_type` with arrays be handled by the fallback, which calls the method below, and further one of the two subsequent ones.
```suggestion
```
Then in the `_evaluate` method, the `p` argument needs to be removed in the `zero(result_type(d, a, b, p))`.

johnnychen94: Much clearer!

dkarrasch:
Fantastic! If you could increment the version number to 0.9.2, then I can merge and release right away.

johnnychen94:
Thanks for taking care of this ❤️

PR #182, opened 2020-10-03
Change Mahalanobis input to AbstractMatrix

### LINKED ISSUE #181: StaticMatrix with Mahalanobis
Hi

I'm trying to call Mahalanobis with a StaticMatrix and get the following:

```
ERROR: LoadError: MethodError: no method matching Distances.Mahalanobis(::StaticArrays.SArray{Tuple{2,2},Float64,2,4})
Closest candidates are:
  Distances.Mahalanobis(::Array{T,2}) where T
```

Would it be okay to open a PR to change the definition to something like that?

```julia
struct Mahalanobis{T} <: Metric
    qmat::AbstractMatrix{T}
end

struct SqMahalanobis{T} <: SemiMetric
    qmat::AbstractMatrix{T}
end
```

rmcaixeta:
This closes #181.
Would be good to accept AbstractMatrix as input to Mahalanobis
Otherwise, it'd not accept a StaticMatrix input
Made a simple adjust and tested here with a StaticMatrix. Working nice

### DIFF
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -1,15 +1,15 @@
 # Mahalanobis distances
 
-struct Mahalanobis{T,M<:AbstractMatrix{T}} <: Metric
+struct Mahalanobis{M<:AbstractMatrix} <: Metric
     qmat::M
 end
 
-struct SqMahalanobis{T,M<:AbstractMatrix{T}} <: SemiMetric
+struct SqMahalanobis{M<:AbstractMatrix} <: SemiMetric
     qmat::M
 end
 
-result_type(::Mahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
-result_type(::SqMahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
+result_type(d::Mahalanobis, ::AbstractVector{T1}, ::AbstractVector{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
+result_type(d::SqMahalanobis, ::AbstractVector{T1}, ::AbstractVector{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
 
 # SqMahalanobis
 
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -8,8 +8,15 @@ struct SqMahalanobis{M<:AbstractMatrix} <: SemiMetric
     qmat::M
 end
 
-result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
-result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
+function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
+    z = oneunit(T1) - oneunit(T2)
+    return typeof(z * oneunit(eltype(d.qmat)) * z)
+end
+
+function result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
+    z = oneunit(T1) - oneunit(T2)
+    return typeof(z * oneunit(eltype(d.qmat)) * z)
+end
 
 # SqMahalanobis
 
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -1,11 +1,11 @@
 # Mahalanobis distances
 
 struct Mahalanobis{T} <: Metric
-    qmat::Matrix{T}
+    qmat::AbstractMatrix{T}
 end
 
 struct SqMahalanobis{T} <: SemiMetric
-    qmat::Matrix{T}
+    qmat::AbstractMatrix{T}
 end
 
 result_type(::Mahalanobis{T}, ::Type, ::Type) where {T} = T
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -9,13 +9,13 @@ struct SqMahalanobis{M<:AbstractMatrix} <: SemiMetric
 end
 
 function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
-    z = oneunit(T1) - oneunit(T2)
-    return typeof(z * oneunit(eltype(d.qmat)) * z)
+    z = zero(T1) - zero(T2)
+    return typeof(sqrt(z * zero(eltype(d.qmat)) * z))
 end
 
 function result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
-    z = oneunit(T1) - oneunit(T2)
-    return typeof(z * oneunit(eltype(d.qmat)) * z)
+    z = zero(T1) - zero(T2)
+    return typeof(z * zero(eltype(d.qmat)) * z)
 end
 
 # SqMahalanobis
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -8,8 +8,8 @@ struct SqMahalanobis{M<:AbstractMatrix} <: SemiMetric
     qmat::M
 end
 
-result_type(d::Mahalanobis, ::AbstractVector{T1}, ::AbstractVector{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
-result_type(d::SqMahalanobis, ::AbstractVector{T1}, ::AbstractVector{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
+result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
+result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
 
 # SqMahalanobis
 
--- a/src/mahalanobis.jl
+++ b/src/mahalanobis.jl
@@ -1,19 +1,19 @@
 # Mahalanobis distances
 
-struct Mahalanobis{T} <: Metric
-    qmat::AbstractMatrix{T}
+struct Mahalanobis{T,M<:AbstractMatrix{T}} <: Metric
+    qmat::M
 end
 
-struct SqMahalanobis{T} <: SemiMetric
-    qmat::AbstractMatrix{T}
+struct SqMahalanobis{T,M<:AbstractMatrix{T}} <: SemiMetric
+    qmat::M
 end
 
-result_type(::Mahalanobis{T}, ::Type, ::Type) where {T} = T
-result_type(::SqMahalanobis{T}, ::Type, ::Type) where {T} = T
+result_type(::Mahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
+result_type(::SqMahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
 
 # SqMahalanobis
 
-function (dist::SqMahalanobis{T})(a::AbstractVector, b::AbstractVector) where {T <: Real}
+function (dist::SqMahalanobis)(a::AbstractVector, b::AbstractVector)
     if length(a) != length(b)
         throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
     end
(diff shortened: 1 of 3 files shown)


src/mahalanobis.jl
```diff
@@ -1,19 +1,19 @@
...
-    qmat::AbstractMatrix{T}
+struct Mahalanobis{T,M<:AbstractMatrix{T}} <: Metric
+    qmat::M
 end
 
-struct SqMahalanobis{T} <: SemiMetric
-    qmat::AbstractMatrix{T}
+struct SqMahalanobis{T,M<:AbstractMatrix{T}} <: SemiMetric
+    qmat::M
 end
```
dkarrasch: These changes are the only necessary.

juliohm: I think we can also eliminate the mentions to `T` in function definitions @dkarrasch given that they are not used in the function body. Also, eliminate the premature `T<:Real` that appears in some of them. Ideally we leave it free to be any type as people may want to use the distance with exotic entry types.

src/mahalanobis.jl
```diff
@@ -1,19 +1,19 @@
...
 
-result_type(::Mahalanobis{T}, ::Type, ::Type) where {T} = T
-result_type(::SqMahalanobis{T}, ::Type, ::Type) where {T} = T
+result_type(::Mahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
+result_type(::SqMahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
 
 # SqMahalanobis
 
-function (dist::SqMahalanobis{T})(a::AbstractVector, b::AbstractVector) where {T <: Real}
+function (dist::SqMahalanobis)(a::AbstractVector, b::AbstractVector)
```
juliohm, line 16: Nice, I like that we are removing the restriction on Real here as well.

juliohm:
Could someone please take a look at this PR? We found an application downstream that requires static allocated matrices for performance.

src/mahalanobis.jl
```diff
@@ -1,19 +1,19 @@
...
-struct SqMahalanobis{T} <: SemiMetric
-    qmat::AbstractMatrix{T}
+struct SqMahalanobis{T,M<:AbstractMatrix{T}} <: SemiMetric
+    qmat::M
 end
 
-result_type(::Mahalanobis{T}, ::Type, ::Type) where {T} = T
-result_type(::SqMahalanobis{T}, ::Type, ::Type) where {T} = T
+result_type(::Mahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
+result_type(::SqMahalanobis{T,M}, ::Type, ::Type) where {T,M} = T
```
dkarrasch, line 12: No need to mention `M` here. Trailing types can be left implicit.

src/mahalanobis.jl
dkarrasch: In fact, these restrictions are "wrong", because they don't apply to the type, right? Formally, we don't have distance computation methods for eltypes other than `<:Real`.

juliohm: Yep. I think we can eliminate the restriction in the code, and let users figure out if a particular implementation actually works with other numbers that aren't real. I think this is one of those antipatterns in Julia where we prematurely restrict the type without a clear benefit.

src/mahalanobis.jl
juliohm: Should this be modified before the PR can be merged @dkarrasch or it can be left that way? I personally find the explicit version with the 2 type parameters better.

dkarrasch: Sure, this is just a style comment. OTOH, it's Julia style not to spell out all parameters. There is one other thing that confuses me. Is it really intentional to have the eltype of the matrix determine the `result_type`? What if you have an integer matrix, but float vectors? Shouldn't the result_type be sufficiently large to hold `one(Ta)*one(T)*one(Tb)[ + one(Ta)*one(T)*one(Tb)]`?

juliohm: That is a good point. I guess #179 also tries to fix this issue, right? The result type should be a combination of all argument types with `promote_type`.

juliohm: @johnnychen94 since you are up-to-date with the linked issue above, what is your suggestion for the `result_type` of Mahalanobis here? Can it just be the following?

```julia
result_type(::Mahalanobis{T,M}, ::AbstractVector{T1}, ::AbstractVector{T2}) where {T,T1,T2} = reduce(promote_type, (T,T1,T2))
```

johnnychen94: No need to wrap it with `reduce` here:

```julia
promote_type(T, T1, T2)
```

would be fine.

> What if you have an integer matrix, but float vectors?

A good example, probably we can also strip off the `T` type in the struct definition, i.e.,

```julia
struct Mahalanobis{M<:AbstractMatrix} <: Metric
    qmat::M
end
```

Both of these should be seen as breaking changes (e.g., memory allocation is sensitive to result type), so I think it's safer to get it in v0.10 releases.

src/mahalanobis.jl
johnnychen94: echo 👍 I like this change

I was quite annoyed with these specific type annotations because the whole JuliaImages world are built on top of `Gray{Float32}` things :P and none of them are Real...

```julia
julia> Gray{Float32} <: Colorant
true

julia> Gray{Float32} <: Real
false
```

src/mahalanobis.jl
juliohm: @rmcaixeta could you please implement the proposed changes? Update the Mahalanobis struct as suggested by @johnnychen94 to contain a single type parameter `M` and update the result_type methods to something like

```Julia
result_type(d::Mahalanobis, ::AbstractVector{T1}, ::AbstractVector{T2}) = promote_type(eltype(d.qmat), T1, T2)
```

dkarrasch: I fixed a mild typo above. Also, tests will need to be adjusted, because currently they are testing if the result type is equal to the eltype of `Q`. I hope the `F64` thing will go through smoothly.

rmcaixeta: Thanks. PR updated

dkarrasch: Wait, I overlooked that. I think it should be
```julia
result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2} = promote_type(eltype(d.qmat), T1, T2)
```
to stay in line with the previous signature.

juliohm: @dkarrasch the last time I checked the internals of Distances.jl the `result_type` function took the distance and two vectors as input. Did that change? The PR was updated correctly by @rmcaixeta assuming that these are still the inputs.

dkarrasch: No. Look at the generic fallback:

https://github.com/JuliaStats/Distances.jl/blob/777f02307d152adc8dedca27df736a61efd975b8/src/generic.jl#L36

For specific metrics, we need to specialize `result_type` for types, not for arrays, not to mention for vectors. This is also the current signature

https://github.com/JuliaStats/Distances.jl/blob/777f02307d152adc8dedca27df736a61efd975b8/src/mahalanobis.jl#L11-L12

which simply ignores the eltypes of the arrays, and takes the eltype of `Q`.

rmcaixeta: Done

dkarrasch: Hmmm, (sorry for being so picky, but I want to get it really right): what if the elements of the vectors are matrices (or colorants or whatever), and (potentially) `mahalanobis` is implemented via 3-arg `dot`? Is the `promote_type` approach gonna work? @johnnychen94 @devmotion 

We use
```julia
function dot(x::AbstractVector, A::AbstractMatrix, y::AbstractVector)
    (axes(x)..., axes(y)...) == axes(A) || throw(DimensionMismatch())
    T = typeof(dot(first(x), first(A), first(y)))
...
```
to determine the result type in 3-arg `dot`. We could go back to the array-based result type determination, like
```julia
result_type(d::Mahalanobis, x::AbstractArray{T1}, y::AbstractArray{T2}) where {T1,T2} = typeof((first(x)-first(y))*first(d.qmat))
```
? I apologize once more for my back and forth, but since we're dropping the strict eltype requirement, we need to have it work for a wild variety of eltypes.

devmotion: IMO it is better to not use `promote_type` but rather a similar approach as in other places currently by defining
```julia
result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2} = typeof(oneunit(eltype(d.qmat)) * oneunit(T1) * oneunit(T2))
```
In contrast to `promote_type`, this will work with unitful numbers as well.

I also don't think one should worry about `a` and `b` being vectors of matrices etc. but rather adopt the convention that the type based definitions of `result_type` *always* operate with the element types. Supporting vectors of matrices, colorants, etc. should then be possible by just defining
```julia
result_type(d::Metric, a::MyCustomType, b::MyCustomType) = result_type(d, get_the_eltype(a), get_the_eltype(b))
```
similar to the existing definitions for `AbstractArray`.

dkarrasch: Sounds good. I changed a detail, `oneunit(T1) - oneunit(T2)`, because this is how it's computed, and this may yield a new type (I'm thinking of `T1 == T2 == Bool`).

devmotion: Ah, thanks for noticing! Actually, to correct my statement about unitful numbers, it should be
```julia
function result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    z = oneunit(T1) - oneunit(T2)
    return typeof(z * oneunit(eltype(d.qmat)) * z)
end
```
since for unitful numbers `typeof(z * z) != typeof(z)`.

johnnychen94: When I introduced this approach in #140, my main focus was that method dispatching becomes way too confusing to write correctly, and it was very easy to introduce method ambiguity. For example, https://github.com/JuliaImages/ImageDistances.jl/issues/28#issuecomment-495010897

Given that Distances.jl computes numerical data, container type (array or colorant) is not needed to infer the result type in cases so there is no need to dispatch on arrays.

rmcaixeta: Nice. @dkarrasch , PR updated with the 2nd correction:

```julia
function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    z = oneunit(T1) - oneunit(T2)
    return typeof(z * oneunit(eltype(d.qmat)) * z)
end

function result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    z = oneunit(T1) - oneunit(T2)
    return typeof(z * oneunit(eltype(d.qmat)) * z)
end
```

src/mahalanobis.jl
```diff
@@ -1,19 +1,26 @@
...
+struct SqMahalanobis{M<:AbstractMatrix} <: SemiMetric
+    qmat::M
 end
 
-result_type(::Mahalanobis{T}, ::Type, ::Type) where {T} = T
-result_type(::SqMahalanobis{T}, ::Type, ::Type) where {T} = T
+function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
+    z = oneunit(T1) - oneunit(T2)
+    return typeof(z * oneunit(eltype(d.qmat)) * z)
+end
```
devmotion, line 14: IMO this is not completely correct, and should be changed to
```suggestion
function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    z = oneunit(T1) - oneunit(T2)
    return typeof(sqrt(z * oneunit(eltype(d.qmat)) * z))
end
```
or
```suggestion
function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    z = zero(T1) - zero(T2)
    return typeof(sqrt(z * zero(eltype(d.qmat)) * z))
end
```
(I noticed that maybe it's more natural to use the additive identity element instead of `oneunit` here)
It could also just be based on the `result_type` for `SqMahalanobis` by writing
```suggestion
function result_type(d::Mahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    return typeof(sqrt(zero(result_type(SqMahalanobis(d.qmat), T1, T2))))
end
```

src/mahalanobis.jl
```diff
@@ -1,19 +1,26 @@
...
+
+function result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
+    z = oneunit(T1) - oneunit(T2)
+    return typeof(z * oneunit(eltype(d.qmat)) * z)
+end
```
devmotion, line 19: As mentioned above, an alternative would be to use `zero`
```suggestion
function result_type(d::SqMahalanobis, ::Type{T1}, ::Type{T2}) where {T1,T2}
    z = zero(T1) - zero(T2)
    return typeof(z * zero(eltype(d.qmat)) * z)
end
```
since it might be more natural and more likely to be defined.

test/test_dists.jl
```diff
@@ -360,9 +360,10 @@ end
         Q = rand(T, length(x), length(x))
         Q = Q * Q'  # make sure Q is positive-definite
         @test sqmahalanobis(x, y, Q) ≈ dot(x - y, Q * (x - y))
-        @test eltype(sqmahalanobis(x, y, Q)) == T
+        z = oneunit(eltype(Q))
+        @test eltype(sqmahalanobis(x, y, Q)) == typeof(z * (oneunit(eltype(x)) - oneunit(eltype(y))) * z)
```
devmotion, line 364: Was this test failing and had to be changed? It seems it should still pass?

rmcaixeta: Thanks for the reviews :+1: 
The tests pass only with  `== T` indeed. There was some mention about reviewing the tests, thought this part might crash. But all is passing as it is. I'll restore the originals

dkarrasch:
Very nice @rmcaixeta! If you could change the version number in the `Project.toml` file to "0.10.0", then I can merge and release right away. 🎉

dkarrasch (approved):

rmcaixeta:
Great. Thanks everyone!

juliohm:
Thank you guys! Looking forward to having this merged and tagged. We have a PR downstream waiting for these changes. 👍

dkarrasch:
Sorry, again, could you please resolve the merge conflict. I don't want to mess up the commits.

johnnychen94:
Never experienced such kind of discussion & collaboartion outside of the Julia community ❤️

rmcaixeta:
> Sorry, again, could you please resolve the merge conflict. I don't want to mess up the commits.

Sure. Guess now it's okay

nalimilan:
It would have been nice to add a test, otherwise `StaticMatrix` support can be broken at any time without anybody noticing. Also now non-`Real` inputs are allowed, but this isn't tested either.

PR #184, opened 2020-10-09
Extend Mahalanobis tests

dkarrasch:


juliohm:
It would be nice to address the use case with static arrays as well as pointed out by @nalimilan

dkarrasch:
Well, that requires us to include a new test dependency. If you want, the `Hermitian(Q)` plays the role of a static matrix, because it's an `AbstractMatrix`, not a `Matrix` anymore. You couldn't construct `Mahalanobis(Hermitian(Q))` before v0.10.

juliohm:
Got it. It makes sense to test with other matrix types that don't add dependencies to the test suite.

nalimilan (approved):
Thanks!

How hard would it be to test an array with a non-`Real` eltype while you're at it?

dkarrasch:
Is `ComplexF64` non-`Real` enough? 😉

PR #185, opened 2020-10-14
relax result_type dist types

johnnychen94:
I'm working on an extension package to Distances to handle massive pointwise calculation where the output is of size at O(m^2*n^2), and I find it could be quite useful to reuse this `result_type` here, so this PR is just relaxing it a bit to accept any kind of function `dist`.

Not very sure if this is a wanted change. I can make up my own `result_type` function there but I guess to get this done here it would be helpful also to other potential usages.

For Distances own purpose, now it is possible to

```julia
julia> result_type(euclidean, Int, Int)
Float64
```

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -26,14 +26,16 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
 # Generic functions
 
 """
-    result_type(dist::PreMetric, Ta::Type, Tb::Type) -> T
-    result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) -> T
+    result_type(dist, Ta::Type, Tb::Type) -> T
+    result_type(dist, a::AbstractArray, b::AbstractArray) -> T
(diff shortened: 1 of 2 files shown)


johnnychen94:
Without feedback in two weeks, I guess this is probably not a good idea to Distances.jl, so I'm going to close it.

dkarrasch:
> For Distances own purpose, now it is possible to
> 
> ```julia
> julia> result_type(euclidean, Int, Int)
> Float64
> ```

I think this is somehow consistent with the types being callable:
```julia
julia> euclidean(2, 3)
1.0

julia> Euclidean()(2, 3)
1.0

julia> result_type(Euclidean(), Int, Int)
Float64

julia> result_type(euclidean, Int, Int)
ERROR: MethodError: no method matching result_type(::typeof(euclidean), ::Type{Int64}, ::Type{Int64})
```
So, I'm in favor.

Issue #187, opened 2020-10-23
Define Distances on Iterators


matthieugomez:
A lot of distances could be defined on any iterator. Would it be possible to relax the signature of `evaluate`? For instance `Hamming` makes sense for strings (e.g. (see https://github.com/matthieugomez/StringDistances.jl/issues/35)) but, I cannot extend the method in [StringDistances.jl](https://github.com/matthieugomez/StringDistances.jl) without type piracy

PR #188, opened 2020-10-26
Added generic function for abstractvectors

### LINKED ISSUE #152: pairwise for arrays of arrays
It would be nice to be able to be able to deal with arrays of arrays as well.
For example : 
```julia
A = [rand(3) for _ in 1:3]
P = pairwise(SqEuclidean(),A)
# Where pairwise(SqEuclidean(),hcat(A,...),dims=2) == P
```
In this case one could of course simply concatenate vectors, but for more complex objects (not necessarily real vectors) that would broaden the use of pairwise!
I could make a PR if you think that makes sense

### LINKED ISSUE #165: pairwise(::PreMetric, ::AbstractVector{<:Real})
To compute the pairwise distance between elements of eg. `x::Vector{Float64}`, one currently has to do something like
```julia
pairwise(d, reshape(x, :, 1); dims=1)
```
Would a PR that wraps the above functionality to make
```julia
pairwise(d, x)
pairwise(d, x, y)
```
do the obvious thing for vectors `x` and `y` be welcome?

theogf:
Add supports for vector of vectors by implementing a generic function.
Fixes #152, closes #165, could be extended to #187 (not sure how to dispatch on Iterators in general, but the current implementation would support it)


theogf:
Bump

theogf:
Bump :)

src/generic.jl
```diff
@@ -120,6 +120,35 @@ function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a::AbstractMatrix)
...
+    n = length(a)
+    size(r) == (n, n) || throw(DimensionMismatch("Incorrect size of r."))
+    @inbounds for j = 1:n
+        for i = (j + 1):n
+            r[i, j] = metric(a[i], a[j])
+        end
+        r[j, j] = 0
+        for i = 1:(j - 1)
+            r[i, j] = r[j, i]   # leveraging the symmetry of SemiMetric
+        end
```
dkarrasch, line 146: How about flipping the two loops? Then you would go through the array in the order of the elements.

test/test_dists.jl
```diff
@@ -504,13 +504,17 @@ function test_pairwise(dist, x, y, T)
...
+        vecy = collect(y[:, i] for i in 1:ny)
         # As earlier, we have small rounding errors in accumulations
         @test pairwise(dist, x, y, dims=2) ≈ rxy
         @test pairwise(dist, x, dims=2) ≈ rxx
         @test pairwise(dist, x, y, dims=2) ≈ rxy
         @test pairwise(dist, x, dims=2) ≈ rxx
         @test pairwise(dist, permutedims(x), permutedims(y), dims=1) ≈ rxy
         @test pairwise(dist, permutedims(x), dims=1) ≈ rxx
+        @test pairwise(dist, vecx, vecy) ≈ rxy
+        @test pairwise(dist, vecx) ≈ rxx
```
dkarrasch, line 517: This would also allow computing the pairwise distance matrix for number vectors, by computing the pairwise scalar distances. This currently errors:
```julia
julia> using Distances

julia> a = rand(5);

julia> pairwise(Euclidean(), a, a)
ERROR: MethodError: no method matching pairwise(::Euclidean, ::Array{Float64,1}, ::Array{Float64,1})
```
Allowing that is only consistent, because we do have, say, `Euclidean()` for both numbers and arrays of numbers. But then we should test it, maybe something like:
```suggestion
        @test pairwise(dist, vecx) ≈ rxx
        vec1 = x[:, 1]; vec2 = y[:, 1]
        @test pairwise(dist, vec1, vec2) ≈ dist.(vec1, permutedims(vec2))
```

dkarrasch (commented):
This looks good to me, except for some minor issues. I should say that this is very hard to tell if it will cause issues with what people may have in their packages, but at least I don't see any.

test/test_dists.jl
dkarrasch: Aha, I see, this was indeed proposed in #165, as you mentioned. 👍

src/generic.jl
theogf: I used the same approach as for matrices as in line 107 but I can switch it again.

src/generic.jl
```diff
@@ -120,6 +120,35 @@ function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a::AbstractMatrix)
...
+    @inbounds for (j, bj) = enumerate(b)
+        for (i, ai) = enumerate(a)
+            r[i, j] = metric(ai, bj)
+        end
+    end
+    r
+end
+
+function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a::AbstractVector)
+    n = length(a)
```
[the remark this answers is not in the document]
theogf: Done!

theogf:
Thanks for the review, I included the changes you suggested. I am not sure what you meant by switching the loop orders. I tried to be consistent with the existing implementation, but it should not matter in the end

dkarrasch:
I meant to organize the loops in such a way that you write into `r` in memory-order, so have the 1:j-1 loop before the j+1:end loop, but otherwise keep what is done within the loops.

It seems the org ran out of Travis credits, so CI doesn't work anymore. We need to move to GithubActions...

theogf:
Ok then I made the right change.

> It seems the org ran out of Travis credits, so CI doesn't work anymore. We need to move to GithubActions...

I can make a PR for that too :)

dkarrasch:
> I can make a PR for that too :)

I think I have something running on my fork. When tests pass there, I'll kick off the PR.

src/generic.jl
```diff
@@ -120,6 +120,37 @@ function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a::AbstractMatrix)
...
+    size(r) == (n, n) || throw(DimensionMismatch("Incorrect size of r."))
+    @inbounds for j = 1:n
+        for i = 1:(j - 1)
+            r[i, j] = metric(a[i], a[j])
+        end
+        r[j, j] = 0
+        for i = (j + 1):n
+            r[i, j] = r[j, i]   # leveraging the symmetry of SemiMetric
+        end
+    end
```
dkarrasch, line 149: I think it should be
```suggestion
        for i = 1:(j - 1)
            r[i, j] = r[j, i]   # leveraging the symmetry of SemiMetric
        end
        r[j, j] = 0
        for i = (j + 1):n
            r[i, j] = metric(a[i], a[j])
        end
    end
```
otherwise you copy from places that you haven't filled yet, if I'm not mistaken.

theogf: Oh my bad, unfortunate copy/paste

dkarrasch:
I think we can do the same loop swap in the matrix version. It's formally the exact same amount of work, but may benefit from better memory access pattern. To the very least, it shouldn't be worse.

theogf:
Tests are failing because of the tests on the `AbstractVector{<:Real}` as you said. I will just remove these tests for now as I guess it deserves a different PR.

src/generic.jl
dkarrasch: Why not use this order of the loops? For each column index `j`, you would start from row `i in 1:j-1`, then fill `r[j,j]`, and then contiguously continue with `i in j+1:n`. I meant to switch the order of the loops, viewed as entire blocks, including their loop body.

theogf: Ok I think I got confused at first but it's more clear now. I will change it.

dkarrasch:
Awesome! Now let's apply the same trick to the good old matrix-version, and I think this is ready to go. For the `Vector{<:Number}` tests, I think we need to restrict to a subset of the metrics, and in particular exclude the weighted metrics (or adjust their parameter vector). Maybe by checking something like `parameters(dist) === nothing`? And the `[Sq]Mahalanobis` stuff failed.

src/generic.jl
```diff
@@ -212,3 +264,18 @@ function pairwise(metric::PreMetric, a::AbstractMatrix;
     r = Matrix{result_type(metric, a, a)}(undef, n, n)
     pairwise!(r, metric, a, dims=dims)
 end
+
+
+function pairwise(metric::PreMetric, a::AbstractVector, b::AbstractVector)
```
dkarrasch, line 269:
```suggestion
function pairwise(metric::PreMetric, a::AbstractVector, b::AbstractVector;
                  dims::Union{Nothing,Integer}=nothing)
    dims = deprecated_dims(dims)
    dims == 2 || throw(ArgumentError("dims must be 2 (got $dims)"))
```

src/generic.jl
```diff
@@ -212,3 +264,18 @@ function pairwise(metric::PreMetric, a::AbstractMatrix;
...
+    m = length(a)
+    n = length(b)
+    r = Matrix{result_type(metric, a, b)}(undef, m, n)
+    pairwise!(r, metric, a, b)
+end
+
+
+function pairwise(metric::PreMetric, a::AbstractVector)
```
dkarrasch, line 277:
```suggestion
function pairwise(metric::PreMetric, a::AbstractVector;
                  dims::Union{Nothing,Integer}=nothing)
    dims = deprecated_dims(dims)
    dims == 2 || throw(ArgumentError("dims must be 2 (got $dims)"))
```

src/generic.jl
```diff
@@ -212,3 +264,18 @@ function pairwise(metric::PreMetric, a::AbstractMatrix;
...
+    n = length(a)
+    r = Matrix{result_type(metric, a, a)}(undef, n, n)
+    pairwise!(r, metric, a)
+end
```
dkarrasch, line 281:
```suggestion
end

```

test/test_dists.jl
```diff
@@ -504,13 +504,17 @@ function test_pairwise(dist, x, y, T)
...
     end
 end
```
dkarrasch, line 519:
```suggestion
        @test pairwise(dist, vecx, vecy, dims=2) ≈ rxy
        @test pairwise(dist, vecx, dims=2) ≈ rxx
    end
end

function test_scalar_pairwise(dist, x, y, T)
    @testset "Scalar pairwise test for $(typeof(dist))" begin
        rxy = dist.(x, permutedims(y))
        rxx = dist.(x, permutedims(x))
        # As earlier, we have small rounding errors in accumulations
        @test pairwise(dist, x, y, dims=2) ≈ rxy
        @test pairwise(dist, x, dims=2) ≈ rxx
        @test_throws ArgumentError pairwise(dist, x, y, dims=1) ≈ rxy
        @test_throws ArgumentError pairwise(dist, x, dims=1) ≈ rxx
        @test pairwise(dist, permutedims(x), permutedims(y), dims=2) ≈ rxy
        @test pairwise(dist, permutedims(x), dims=2) ≈ rxx
        @test_throws DimensionMismatch pairwise(dist, permutedims(x), permutedims(y), dims=1)
    end
end

@testset "scalar pairwise metrics on $T" for T in (Float64, F64)
    m = 5
    n = 8
    nx = 6
    ny = 8

    x = rand(T, nx)
    y = rand(T, ny)

    a = rand(1:3, nx)
    b = rand(1:3, ny)

    test_scalar_pairwise(SqEuclidean(), x, y, T)
    test_scalar_pairwise(Euclidean(), x, y, T)
    test_scalar_pairwise(Cityblock(), x, y, T)
    test_scalar_pairwise(TotalVariation(), x, y, T)
    test_scalar_pairwise(Chebyshev(), x, y, T)
    test_scalar_pairwise(Minkowski(2.5), x, y, T)
    test_scalar_pairwise(Hamming(), a, b, T)
    test_scalar_pairwise(CosineDist(), x, y, T)
    test_scalar_pairwise(CosineDist(), a, b, T)
    test_scalar_pairwise(ChiSqDist(), x, y, T)
    test_scalar_pairwise(KLDivergence(), x, y, T)
    test_scalar_pairwise(JSDivergence(), x, y, T)
    test_scalar_pairwise(BrayCurtis(), x, y, T)

    w = rand(1, 1)
    test_scalar_pairwise(WeightedSqEuclidean(w), x, y, T)
    test_scalar_pairwise(WeightedEuclidean(w), x, y, T)
    test_scalar_pairwise(WeightedCityblock(w), x, y, T)
    test_scalar_pairwise(WeightedMinkowski(w, 2.5), x, y, T)
    test_scalar_pairwise(WeightedHamming(w), a, b, T)
    test_scalar_pairwise(PeriodicEuclidean(w), x, y, T)
end
```

dkarrasch (commented):
I have added scalar tests and the required `dims` keyword to `pairwise`. I suggest to replace the three occurrences of `transpose` in `generic.jl` by `permutedims`, since the first acts recursively, but the `dims=1` vs `dims=2` distinction is only about the orientation of the outermost array.

src/generic.jl
theogf: I am not sure what is the point of giving a `dims` argument. It is meaningless in this case.

dkarrasch: Sure, but that's the API, AFAIU. In the matrix case, you even get a warning when you don't provide the `dims` keyword, so it's mandatory. And in the vector case, you included it correctly in `pairwise!`  yourself. I think we should include some `dims` handling there, too. But in any case we need to be consistent.

theogf: Oh fair point. But in this case I would remove the `dims` in `pairwise!` then. I don't think one needs the exact same API for Vectors and Matrices. This could even be a source of bugs.
Just to refer to KernelFunctions, we provide an API to both but only the Matrix API takes a `dims` argument

dkarrasch: I see. We should ask for @nalimilan's advice then, because he introduced that mandatory keyword argument for consistency with some other packages from the Statistics ecosystem.

BTW, I also noticed that dims-warning when running the tests from this PR.

src/Distances.jl
```diff
@@ -97,6 +97,13 @@ export
     rmsd,
     nrmsd
 
+if VERSION < v"1.2-"
+    import Base: has_offset_axes
+    require_one_based_indexing(A...) = !has_offset_axes(A...) || throw(ArgumentError("offset arrays are not supported but got an array with index other than 1"))
```
nalimilan, line 102:
```suggestion
    require_one_based_indexing(A...) =
        !has_offset_axes(A...) ||
            throw(ArgumentError("offset arrays are not supported but got an array with index other than 1"))
```

src/generic.jl
```diff
@@ -33,8 +33,8 @@ Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
 result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
-
+result_type(dist::PreMetric, a::AbstractArray{<:Number}, b::AbstractArray{<:Number}) = result_type(dist, eltype(a), eltype(b))
+result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(first(a)), eltype(first(b)))
```
nalimilan, line 37: This is going to be problematic for e.g. `AbstractArray{Union{T, Missing}}`. Can't we require inputs to have precise enough eltypes?

src/generic.jl
```diff
@@ -109,17 +109,48 @@ function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a::AbstractMatrix)
     size(r) == (n, n) || throw(DimensionMismatch("Incorrect size of r."))
     @inbounds for j = 1:n
         aj = view(a, :, j)
+        for i = 1:(j - 1)
+            r[i, j] = r[j, i]   # leveraging the symmetry of SemiMetric
+        end
```
nalimilan, line 114: There are two things I don't understand:
- How did this work before? Don't we already do this somewhere else?
- Why copy *before* filling the other part?

src/generic.jl
```diff
@@ -109,17 +109,48 @@ function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a::AbstractMatrix)
...
         for i = (j + 1):n
             r[i, j] = metric(view(a, :, i), aj)
         end
-        r[j, j] = 0
+    end
+    r
+end
+
+function _pairwise!(r::AbstractMatrix, metric::PreMetric,
+                    a::AbstractVector, b::AbstractVector=a)
```
nalimilan, line 124: With the default value the 3-arg method is defined twice:
```suggestion
function _pairwise!(r::AbstractMatrix, metric::PreMetric,
                    a::AbstractVector, b::AbstractVector)
```

src/generic.jl
```diff
@@ -185,8 +218,27 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
     end
 end
 
+function pairwise!(r::AbstractMatrix, metric::PreMetric,
+    a::AbstractVector, b::AbstractVector;
+    dims::Union{Nothing,Integer}=nothing)
```
nalimilan, line 223: Align arguments.

src/generic.jl
```diff
@@ -185,8 +218,27 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
...
+    na = length(a)
+    nb = length(b)
+    size(r) == (na, nb) ||
+        throw(DimensionMismatch("Incorrect size of r (got $(size(r)), expected $((na, nb)))."))
+    _pairwise!(r, metric, a, b)
+end
+
+function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractVector;
+    dims::Union{Nothing,Integer}=nothing)
```
nalimilan, line 232: Here too.

src/generic.jl
nalimilan: The idea with the mandatory `dims` argument is that we need to change the default to become consistent with Statistics in the stdlib (e.g. `cor`). So until we do that we have to require people to pass it explicitly so that their code doesn't break the day we make the switch. But here I agree it doesn't make any sense to support it since there are no rows and columns, only vectors of vectors.

src/generic.jl
dkarrasch: Something like keeping the old `result_type` method for arrays, and add
```julia
result_type(dist::PreMetric, a::AbstractArray{<:AbstractArray}, b::AbstractArray{<:AbstractArray}) = result_type(dist, eltype(eltype(a)), eltype(eltype(b)))
```
?

src/generic.jl
dkarrasch: How it worked before: let's say, we want to fill a 3x3 matrix symmetrically, then the old write pattern was like this:
```
 1  6  8
 2  4  9
 3  5  7
```
In 1, 4, 7 we simply fill with a zero. In 2, 3, 5 we compute new distances, in 6, 8 and 9 we copy the values from 2, 3 and 5, respectively, i.e., values computed in earlier iterations.

How it's working here:
```
 1  4  7
 2  5  8
 3  6  9
```
In 1, 5, 9 we simply fill with a zero. In 2, 3, 6 we compute new distances, in 4, 7 and 8 we copy the values from 2, 3 and 6, respectively.

Why do that? Because it fills `r` in memory order, and the result is exactly the same.

nalimilan: Ah OK this relies on the fact that the particular entry that we copy has already been set, even if that's not the case for all cells. Clever.

nalimilan: Also the reason why I was confused is that the diff makes it look like the function didn't have the first loop if you don't look very carefully. :-)

dkarrasch: True, it took me a couple of minutes to digest the diff, too.

src/generic.jl
theogf: Yes because it goes then to the specialized method `_pairwise(r, metric, a)` which takes advantage of the symmetry

nalimilan: What I mean is that if you keep the default value `b=a` you define `_pairwise(r, metric, a)` twice.

theogf: Oh right! Sorry. Well actually it's also true for the AbstractMatrix case... I will correct both

theogf: Ok figured out why it's like this. The one argument method is only for `SemiMetric`... I switched it back

theogf:
@dkarrasch @nalimilan Ok I think all issues are solved now.

src/metrics.jl
```diff
@@ -489,6 +489,8 @@ js_divergence(a::AbstractArray, b::AbstractArray) = JSDivergence()(a, b)
 
 result_type(dist::SpanNormDist, a::AbstractArray, b::AbstractArray) =
      typeof(eval_op(dist, oneunit(eltype(a)), oneunit(eltype(b))))
+result_type(dist::SpanNormDist, a::AbstractArray{<:AbstractArray}, b::AbstractArray{<:AbstractArray}) =
+     typeof(eval_op(dist, oneunit(eltype(first(a))), oneunit(eltype(first(b)))))
```
dkarrasch, line 493:
```suggestion
     typeof(eval_op(dist, oneunit(eltype(eltype(a))), oneunit(eltype(eltype(b)))))
```

dkarrasch (commented):
Two small things, and then I think this is ready.

@nalimilan I wonder if it helps FreqTables.jl if we made these `pairwise` methods fully generic for iterators, and removed the type constraints on `a` and `b`. Then the `Distances.jl` pairwise method would act `metric::PreMetric`.

src/metrics.jl
nalimilan: Maybe we should rather override `result_type(dist::SpanNormDist, a::Type, b::Type)` and have global generic methods that call this? It sounds suboptimal to redefine methods for different inputs types.

dkarrasch:
The more I think about it, the more I believe this really is not about `AbstractVector` methods, but about iterable objects that contain data points between which we want to compute distances. I included iterator-based `pairwise` and `colwise` methods now in #194. My `result_type` implementation generically handles arrays of arrays (or arrays), but avoids things like https://github.com/JuliaStats/Distances.jl/pull/188#discussion_r541088303. I guess it kind of supersedes this PR (I also included that scalar pairwise tests).

PR #191, opened 2020-12-03
Move CI to GitHubActions

dkarrasch:
I took this from `NearestNeighbors.jl`, but removed `cache`-related stuff, that I've not seen elsewhere. @KristofferC, can you please take a look?


.github/workflows/CI.yml
```diff
@@ -0,0 +1,38 @@
...
+          - x64
+    steps:
+      - uses: actions/checkout@v2
+      - uses: julia-actions/setup-julia@v1
+        with:
+          version: ${{ matrix.version }}
+          arch: ${{ matrix.arch }}
+      - uses: julia-actions/julia-buildpkg@v1
+      - uses: julia-actions/julia-runtest@v1
+      - uses: julia-actions/julia-processcoverage@v1
```
KristofferC, line 35: Are both `processcoverage` and the `codecov` action needed?

KristofferC:
> but removed cache-related stuff, that I've not seen elsewhere

I just took it from somewhere else

.github/workflows/CI.yml
dkarrasch: Yes, you can see in the first commit that the codecov action was complaining about a missing file.

KristofferC: Alright, guess I also need to add that to NearestNeighbors.jl heh.

KristofferC (approved):
Can delete the travis file as well?

.github/workflows/CI.yml
```diff
@@ -0,0 +1,38 @@
...
+    branches:
+      - master
+  push:
+    branches:
+      - master
+    tags: '*'
+jobs:
+  test:
+    name: Julia ${{ matrix.version }} - ${{ matrix.os }} - ${{ matrix.arch }} - ${{ github.event_name }}
+    runs-on: ${{ matrix.os }}
```
theogf, line 13:
```suggestion
    runs-on: ${{ matrix.os }}
    continue-on-error: ${{ matrix.version == 'nightly' }}
``
This will allow failure on nightly

theogf:
I know it's been merged already but you might want to change the build badge on the README by 
CI

dkarrasch:
Good point. Let's include that into your ongoing PR. I don't think I can write into this repo directly, and I wouldn't want to create another PR just for that one line.

PR #192, opened 2020-12-04
Relax type signature for `result_type` and allow it to act on numbers

### LINKED ISSUE #190: `result_type` acts on objects for arrays but on types otherwise
https://github.com/JuliaStats/Distances.jl/blob/f97709ff6e04fc290dd090cd0e541d7b9b052475/src/generic.jl#L29-L30

This seems a bit inconsistent, because it makes it hard to use `result_type` in generic code where you could be computing distances between arrays or between scalars. Maybe a fallback like `result_type(metric, a::Number, b::Number) = result_type(metric, typeof(a), typeof(b))` could be added?

I can make a PR if this is acceptable.

ericphanson:
Redo of #185 along with defining `result_type` for `Number`s so that `result_type` can act on objects generically as discussed in #190.

Closes #190

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -33,10 +33,11 @@ Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
 result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
-# for outside extensibility to other packages
-result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b)))
-result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
+# Promote Arays and Numbers to types
+result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(dist, a::Number, b::Number) = result_type(dist, typeof(a), typeof(b))
 
 # Generic column-wise evaluation
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -26,14 +26,16 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
 # Generic functions
 
 """
-    result_type(dist::PreMetric, Ta::Type, Tb::Type) -> T
-    result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) -> T
+    result_type(dist, Ta::Type, Tb::Type) -> T
+    result_type(dist, a::AbstractArray, b::AbstractArray) -> T
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
-result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
+# for outside extensibility to other packages
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b)))
+result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
 
 
 # Generic column-wise evaluation
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -36,7 +36,6 @@ data `a` and `b`.
 result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
 result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
-# Promote Arrays and Numbers to types
 result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
 result_type(dist, a::Number, b::Number) = result_type(dist, typeof(a), typeof(b))
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -28,6 +28,7 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
 """
     result_type(dist, Ta::Type, Tb::Type) -> T
     result_type(dist, a::AbstractArray, b::AbstractArray) -> T
+    result_type(dist, a::Number, b::Number) -> T
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -36,7 +36,7 @@ data `a` and `b`.
 result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
 result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
-# Promote Arays and Numbers to types
+# Promote Arrays and Numbers to types
 result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
 result_type(dist, a::Number, b::Number) = result_type(dist, typeof(a), typeof(b))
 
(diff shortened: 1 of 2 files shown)


src/generic.jl
```diff
@@ -26,15 +26,19 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
-result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
+# Promote Arays and Numbers to types
```
dkarrasch, line 39:
```suggestion
# Promote Arrays and Numbers to types
```

dkarrasch (approved):
LGTM, except for the little typo.

johnnychen94 (approved):

ericphanson:
Bump

johnnychen94:
I believe this one is covered by #194 now

ericphanson:
Ah indeed, I missed that. Ok, should I close this PR?

dkarrasch:
I didn't mean to steal anything, I was just anticipating these changes and copied them. We should try to get feedback by somebody else before merging this. @nalimilan?

src/generic.jl
```diff
@@ -26,15 +26,19 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
-result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
+# Promote Arrays and Numbers to types
```
nalimilan, line 39: This is very different from promotion. Maybe no comment is needed.
```suggestion
```

src/generic.jl
```diff
@@ -26,15 +26,19 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
+    result_type(dist, a::AbstractArray, b::AbstractArray) -> T
+    result_type(dist, a::Number, b::Number) -> T
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
-result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
```
nalimilan, line 37: Are we sure we want to preserve units if inputs have them for all kinds of distances?

nalimilan (commented):
TBH I find it a bit weird to add unexported definitions that aren't used at all by the package. Does https://github.com/JuliaStats/Distances.jl/pull/194 use them at least?

ericphanson:
> TBH I find it a bit weird to add unexported definitions that aren't used at all by the package

do you mean that you consider `result_type` internal and not part of the public API? The reason I want this change is because it's used in downstream code (ref https://github.com/JuliaStats/Distances.jl/issues/190). I can instead make a PR to NearestNeighborDescent if it should be using something other than `Distances.result_type`, e.g. its own vendored copy with this method.

(Relatedly, since Distances.jl doesn't have docs, it's not clear what constitutes the public API-- only exported functions, or functions with docstrings, or some mix thereof?)

src/generic.jl
ericphanson: I'm not sure; I guess that's a question for @johnnychen94, since the method came from #185. To me the "all kinds of distances" isn't super worrying since any distance could just provide a more specialized method, but to be honest I'm not super confident in the interplay between units and distances and maybe that could be tedious or painful.

johnnychen94: Yes, as we've discussed in the original issue #190, this change gives other packages the ability to reuse the name `result_type` without introducing type piracy. Because `Distances` defines `result_type`, the fallback should be defined in `Distances`.

johnnychen94: To be clear, `result_type` is exported by `Distances`.

https://github.com/JuliaStats/Distances.jl/blob/e50899d2f3814c5a5adae74d919782ff0d133405/src/Distances.jl#L13

ericphanson: ah good point, it is exported!

I think @nalimilan's question here was about `oneunit`, any thoughts on that @johnnychen94?

> Are we sure we want to preserve units if inputs have them for all kinds of distances?

johnnychen94: I'm not sure if I understand the question clearly and I don't know what that question intends to suggest.

---

Firstly, I don't know why the fallback is `Float64` for `PreMetric`, it's probably written down when `result_type` wasn't exported and the main consideration is probably arithmetic performance inside Distances only.

FixedPointNumbers has introduced a better strategy `floattype(x)` to rewrite that `N0f8`( normalized `UInt8`) should get promoted to `Float32` instead of `Float16`. I think Distances should follow the same pattern and should not hard code the result type into `Float64`. People don't like changes, so it's okay to keep the internal `Distances` unchanged.

We're not changing `Distances` internal implementation here. As is said, we're adding a fallback that will be overridden by all the internal distances' implementation.

I think for any general function `f` (not limited to Distances itself), calling `f` on the units is a more generic fallback. Let me give two examples here:

For `Unitful`:

```julia
julia> using Unitful, Distances

julia> x = fill(999u"mm", 4, 4);

julia> y = fill(1u"m", 4, 4);

julia> cityblock(x, y)
2//125 m

julia> result_type(Cityblock(), x, y)
Quantity{Rational{Int64},𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}

julia> result_type(cityblock, x, y) # undefined but shouldn't be Float64
```

Also for JuliaImages, `ColorVectorSpaces` has defined a lot of basic operations for `Gray` and `RGB`, so it makes sense to let `result_type` be `Gray`/`RGB` if that's what they are supposed to be.

```julia
julia> using ImageCore, ColorVectorSpace

julia> x = rand(Gray{N0f8}, 4, 4);

julia> -(oneunit(eltype(x)), oneunit(eltype(x)))
Gray{N0f8}(0.0)
```

nalimilan: OK. Actually I hadn't realized that `oneunit` is already used by existing methods:
https://github.com/JuliaStats/Distances.jl/blob/e50899d2f3814c5a5adae74d919782ff0d133405/src/metrics.jl#L216-L221

`Float64` is indeed not a great default. Do you know whether it's actually used in Distances ? Maybe we could remove the fallback method and always rely on `oneunit`.

nalimilan:
> do you mean that you consider `result_type` internal and not part of the public API? The reason I want this change is because it's used in downstream code (ref #190). I can instead make a PR to NearestNeighborDescent if it should be using something other than `Distances.result_type`, e.g. its own vendored copy with this method.

Sorry, I hadn't realized it was exported. I considered it as the equivalent of `Base.promote_op`, which is something nobody is proud of but we haven't been able to replace yet. (In general, as `?Base.promote_op` says, the recommended approach in Julia is to avoid relying on the eltype but instead of use the actual type of elements, and widen the type if necessary, like `map` and `collect` do. But in practice that's quite complex to implement, and Distances predates that approach.)

> (Relatedly, since Distances.jl doesn't have docs, it's not clear what constitutes the public API-- only exported functions, or functions with docstrings, or some mix thereof?)

Exported functions are definitely part of the API. It's less clear for unexported functions that have docstrings: in that case I'd say it depends on what the manual/README says.

src/generic.jl
johnnychen94: By removing this fallback, and observing the test errors, it tells me that only `BhattacharyyaDist`, `HellingerDist` and `Bregman` requires this fallback for its `colwise` implementation. Currently, it makes sense to me to remove this fallback (probably in another PR).

nalimilan: Thanks for checking. I think it would be better to try removing the fallback in this PR. Otherwise the new method wouldn't actually be used for `PreMetric` so it would be misleading to call that `Distances.result_type`!

ericphanson:
Thanks all for the review and discussion; happily closing in favor of https://github.com/JuliaStats/Distances.jl/pull/194.

Issue #177, opened 2020-09-16
Euclidean on cartesian indeces


Dale-Black:
Is there a way to compute the euclidean distance between arrays that are of the type CartesianIndex?

dkarrasch:
Do you mean something like this?
```julia
julia> A = reshape(collect(1:9), 3, 3)
3×3 Array{Int64,2}:
 1  4  7
 2  5  8
 3  6  9

julia> inds1 = findall(iseven, A)
4-element Array{CartesianIndex{2},1}:
 CartesianIndex(2, 1)
 CartesianIndex(1, 2)
 CartesianIndex(3, 2)
 CartesianIndex(2, 3)

julia> inds2 = findall(isodd, A)
5-element Array{CartesianIndex{2},1}:
 CartesianIndex(1, 1)
 CartesianIndex(3, 1)
 CartesianIndex(2, 2)
 CartesianIndex(1, 3)
 CartesianIndex(3, 3)

julia> pairwise(SqEuclidean(), inds1, inds2)
4×5 Array{Int64,2}:
 1  1  1  5  5
 1  5  1  1  5
 5  1  1  5  1
 5  5  1  1  1

julia> euclidean(inds1[1], inds1[1]) === 0.0
true
```
I have added support for this special, noniterable type in #194.

nalimilan:
Could you explain why this would be useful? Since `CartesianIndex` isn't iterable (on purpose), this would require specific support in Distances.

Dale-Black:
I think what @dkarrasch implemented is what I wanted. I would have to go back to this specific project to find out for sure but from what I can tell that solves my use case

nalimilan:
Yes but we decided to revert that change until we have a clearer view of why this would be needed.

dkarrasch:
@nalimilan convinced me that supporting `CartesianIndex`es directly (at least in the context of #194) is awkward, because it targets iterable objects, and `CartesianIndex`es are intentionally non-iterable. I believe the correct way of handling it is to pass the responsibility to the callee, asking her/him to pass iterable objects to `Distances.jl`. In the context here, this would mean to convert `CartesianIndex`es to tuples via `Tuple(cartind)`, as suggested by the error message. Once you have that, the new iterator-based implementation will work just fast and fine.

PR #194, opened 2020-12-05
Allow generators and iterators

### LINKED ISSUE #152: pairwise for arrays of arrays

### LINKED ISSUE #162: Why don't you support NTuples?
```julia
julia> using Distances

julia> euclidean((1,2), (2,3))
ERROR: MethodError: no method matching euclidean(::Tuple{Int64,Int64}, ::Tuple{Int64,Int64})
Stacktrace:
 [1] top-level scope at none:0
```
I am wondering what is behind the decision of not supporting `NTuples`...?

### LINKED ISSUE #165: pairwise(::PreMetric, ::AbstractVector{<:Real})

### LINKED ISSUE #187: Define Distances on Iterators

### LINKED ISSUE #190: `result_type` acts on objects for arrays but on types otherwise

dkarrasch:
This adds support for iterable objects as arguments to `evaluate`,(EDIT: and `pairwise` and `colwise`). ~~I haven't touched pair- and colwise stuff, because that is (partially/completely?) addressed by #188.~~

Closes #187. Closes #162. Closes #165. Closes #152. Closes #190. Closes #192. Closes #188.

### DIFF
--- a/src/bhattacharyya.jl
+++ b/src/bhattacharyya.jl
@@ -40,7 +40,7 @@ end
     asum = zero(Ta)
     bsum = zero(Tb)
 
-    @simd for i = 1:length(a)
+    @simd for i in eachindex(a, b)
         @inbounds ai = a[i]
         @inbounds bi = b[i]
         sqab += sqrt(ai * bi)
--- a/src/bhattacharyya.jl
+++ b/src/bhattacharyya.jl
@@ -9,46 +9,45 @@ struct HellingerDist <: Metric end
 
 # Bhattacharyya coefficient
 
-function bhattacharyya_coeff(a::AbstractVector{T}, b::AbstractVector{T}) where {T <: Number}
-    if length(a) != length(b)
-        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
-    end
-
-    n = length(a)
-    sqab = zero(typeof(sqrt(zero(T))))
-    # We must normalize since we cannot assume that the vectors are normalized to probability vectors.
-    asum = zero(T)
-    bsum = zero(T)
-
-    @simd for i = 1:n
-        @inbounds ai = a[i]
-        @inbounds bi = b[i]
-        sqab += sqrt(ai * bi)
-        asum += ai
-        bsum += bi
-    end
-
-    sqab / sqrt(asum * bsum)
-end
 function bhattacharyya_coeff(a, b)
-    T = typeof(sqrt(zero(promote_type(_eltype(a), _eltype(b)))))
     n = length(a)
     if n != length(b)
         throw(DimensionMismatch("first argument has length $n which does not match the length of the second, $(length(b))."))
     end
-
-    sqab = zero(T)
+    sqab, asum, bsum = _bhattacharyya_coeff(a, b)
     # We must normalize since we cannot assume that the vectors are normalized to probability vectors.
-    asum = zero(T)
-    bsum = zero(T)
+    return sqab / sqrt(asum * bsum)
+end
+
+@inline function _bhattacharyya_coeff(a, b)
+    Ta = _eltype(a)
+    Tb = _eltype(b)
+    T = typeof(sqrt(zero(promote_type(Ta, Tb))))
+    sqab = zero(T)
+    asum = zero(Ta)
+    bsum = zero(Tb)
 
     for (ai, bi) in zip(a, b)
         sqab += sqrt(ai * bi)
         asum += ai
         bsum += bi
     end
+    return sqab, asum, bsum
+end
+@inline function _bhattacharyya_coeff(a::AbstractVector{Ta}, b::AbstractVector{Tb}) where {Ta<:Number,Tb<:Number}
+    T = typeof(sqrt(oneunit(Ta)*oneunit(Tb)))
+    sqab = zero(T)
+    asum = zero(Ta)
+    bsum = zero(Tb)
 
-    return sqab / sqrt(asum * bsum)
+    @simd for i = 1:length(a)
+        @inbounds ai = a[i]
+        @inbounds bi = b[i]
+        sqab += sqrt(ai * bi)
+        asum += ai
+        bsum += bi
+    end
+    return sqab, asum, bsum
 end
 
 # Faster pair- and column-wise versions TBD...
@@ -56,10 +55,8 @@ end
 
 # Bhattacharyya distance
 (::BhattacharyyaDist)(a, b) = -log(bhattacharyya_coeff(a, b))
-(::BhattacharyyaDist)(::Number, ::Number) = throw("Bhattacharyya distance cannot be calculated for scalars")
 bhattacharyya(a, b) = BhattacharyyaDist()(a, b)
 
 # Hellinger distance
 (::HellingerDist)(a, b) = sqrt(1 - bhattacharyya_coeff(a, b))
-(::HellingerDist)(::Number, ::Number) = throw("Hellinger distance cannot be calculated for scalars")
 hellinger(a, b) = HellingerDist()(a, b)
--- a/src/bhattacharyya.jl
+++ b/src/bhattacharyya.jl
@@ -15,7 +15,7 @@ function bhattacharyya_coeff(a::AbstractVector{T}, b::AbstractVector{T}) where {
     end
 
     n = length(a)
-    sqab = zero(T)
+    sqab = zero(typeof(sqrt(zero(T))))
     # We must normalize since we cannot assume that the vectors are normalized to probability vectors.
     asum = zero(T)
     bsum = zero(T)
@@ -30,18 +30,36 @@ function bhattacharyya_coeff(a::AbstractVector{T}, b::AbstractVector{T}) where {
 
     sqab / sqrt(asum * bsum)
 end
+function bhattacharyya_coeff(a, b)
+    T = typeof(sqrt(zero(promote_type(_eltype(a), _eltype(b)))))
+    n = length(a)
+    if n != length(b)
+        throw(DimensionMismatch("first argument has length $n which does not match the length of the second, $(length(b))."))
+    end
 
-bhattacharyya_coeff(a::T, b::T) where {T <: Number} = throw("Bhattacharyya coefficient cannot be calculated for scalars")
+    sqab = zero(T)
+    # We must normalize since we cannot assume that the vectors are normalized to probability vectors.
+    asum = zero(T)
+    bsum = zero(T)
+
+    for (ai, bi) in zip(a, b)
+        sqab += sqrt(ai * bi)
+        asum += ai
+        bsum += bi
+    end
+
+    return sqab / sqrt(asum * bsum)
+end
 
 # Faster pair- and column-wise versions TBD...
 
 
 # Bhattacharyya distance
-(::BhattacharyyaDist)(a::AbstractVector{T}, b::AbstractVector{T}) where {T <: Number} = -log(bhattacharyya_coeff(a, b))
-(::BhattacharyyaDist)(a::T, b::T) where {T <: Number} = throw("Bhattacharyya distance cannot be calculated for scalars")
+(::BhattacharyyaDist)(a, b) = -log(bhattacharyya_coeff(a, b))
+(::BhattacharyyaDist)(::Number, ::Number) = throw("Bhattacharyya distance cannot be calculated for scalars")
 bhattacharyya(a, b) = BhattacharyyaDist()(a, b)
 
 # Hellinger distance
-(::HellingerDist)(a::AbstractVector{T}, b::AbstractVector{T}) where {T <: Number} = sqrt(1 - bhattacharyya_coeff(a, b))
-(::HellingerDist)(a::T, b::T) where {T <: Number} = throw("Hellinger distance cannot be calculated for scalars")
+(::HellingerDist)(a, b) = sqrt(1 - bhattacharyya_coeff(a, b))
+(::HellingerDist)(::Number, ::Number) = throw("Hellinger distance cannot be calculated for scalars")
 hellinger(a, b) = HellingerDist()(a, b)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -29,8 +29,8 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
     result_type(dist, Ta::Type, Tb::Type) -> T
     result_type(dist, a, b) -> T
 
-Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
-data `a` and `b`.
+Infer the result type of metric `dist` with input types `Ta` and `Tb`, or element types
+of iterators `a` and `b`.
 """
 result_type(dist, a, b) = result_type(dist, _eltype(a), _eltype(b))
 result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
@@ -48,7 +48,7 @@ _eltype(T::Type) = T
 # Generic column-wise evaluation
 
 """
-    colwise!(r::AbstractMatrix, metric::PreMetric, a, b)
+    colwise!(r::AbstractArray, metric::PreMetric, a, b)
 
 Compute distances between corresponding elements of the iterable collections
 `a` and `b` according to distance `metric`, and store the result in `r`.
@@ -88,8 +88,8 @@ function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::Abs
 end
 
 """
-    colwise!(r::AbstractMatrix, metric::PreMetric,
-              a::AbstractMatrix, b::AbstractMatrix)
+    colwise!(r::AbstractArray, metric::PreMetric,
+             a::AbstractMatrix, b::AbstractMatrix)
 
 Compute distances between each corresponding columns of `a` and `b` according
 to distance `metric`, and store the result in `r`. Exactly one of `a` or `b`
@@ -110,13 +110,12 @@ function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::Abs
 end
 
 """
-    colwise(r::AbstractMatrix, metric::PreMetric, a, b)
+    colwise(metric::PreMetric, a, b)
 
 Compute distances between corresponding elements of the iterable collections
 `a` and `b` according to distance `metric`.
 
-`a` and `b` must have the same number of elements, `r` must be a vector of length
-`length(a) == length(b)`.
+`a` and `b` must have the same number of elements (`length(a) == length(b)`).
 """
 function colwise(metric::PreMetric, a, b)
     n = get_common_length(a, b)
@@ -125,8 +124,7 @@ function colwise(metric::PreMetric, a, b)
 end
 
 """
-    colwise(r::AbstractMatrix, metric::PreMetric,
-              a::AbstractMatrix, b::AbstractMatrix)
+    colwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix)
 
 Compute distances between each corresponding columns of `a` and `b` according
 to distance `metric`. Exactly one of `a` or `b` can be a vector, in which case
@@ -282,8 +280,7 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
 end
 
 """
-    pairwise!(r::AbstractMatrix, metric::PreMetric, a, b)
-    pairwise!(r::AbstractMatrix, metric::PreMetric, a)
+    pairwise!(r::AbstractMatrix, metric::PreMetric, a, b=a)
 
 Compute distances between each element of collection `a` and each element of
 collection `b` according to distance `metric`, and store the result in `r`.
@@ -296,7 +293,6 @@ pairwise!(r::AbstractMatrix, metric::PreMetric, a) = _pairwise!(r, metric, a)
 
 """
     pairwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix=a; dims)
-    pairwise(metric::PreMetric, a::AbstractMatrix; dims)
 
 Compute distances between each pair of rows (if `dims=1`) or columns (if `dims=2`)
 in `a` and `b` according to distance `metric`. If a single matrix `a` is provided,
@@ -324,8 +320,7 @@ function pairwise(metric::PreMetric, a::AbstractMatrix;
 end
 
 """
-    pairwise(metric::PreMetric, a, b)
-    pairwise(metric::PreMetric, a)
+    pairwise(metric::PreMetric, a, b=a)
 
 Compute distances between each element of collection `a` and each element of
 collection `b` according to distance `metric`. If a single iterable `a` is
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -258,7 +258,7 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric,
     size(r) == (na, nb) ||
         throw(DimensionMismatch("Incorrect size of r (got $(size(r)), expected $((na, nb)))."))
     if dims == 1
-        _pairwise!(r, metric, transpose(a), transpose(b))
+        _pairwise!(r, metric, permutedims(a), permutedims(b))
     else
         _pairwise!(r, metric, a, b)
     end
@@ -276,7 +276,7 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
     size(r) == (n, n) ||
         throw(DimensionMismatch("Incorrect size of r (got $(size(r)), expected $((n, n)))."))
     if dims == 1
-        _pairwise!(r, metric, transpose(a))
+        _pairwise!(r, metric, permutedims(a))
     else
         _pairwise!(r, metric, a)
     end
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -50,7 +50,7 @@ __eltype(::Base.EltypeUnknown, a) = _eltype(typeof(first(a)))
 Compute distances between corresponding elements of the iterable collections
 `a` and `b` according to distance `metric`, and store the result in `r`.
 
-`a` and `b` must have the same number of elements, `r` must be a vector of length
+`a` and `b` must have the same number of elements, `r` must be an array of length
 `length(a) == length(b)`.
 """
 function colwise!(r::AbstractArray, metric::PreMetric, a, b)
@@ -87,6 +87,10 @@ end
 """
     colwise!(r::AbstractArray, metric::PreMetric,
              a::AbstractMatrix, b::AbstractMatrix)
+    colwise!(r::AbstractArray, metric::PreMetric,
+             a::AbstractVector, b::AbstractMatrix)
+    colwise!(r::AbstractArray, metric::PreMetric,
+             a::AbstractMatrix, b::AbstractVector)
 
 Compute distances between each corresponding columns of `a` and `b` according
 to distance `metric`, and store the result in `r`. Exactly one of `a` or `b`
@@ -94,7 +98,11 @@ can be a vector, in which case the distance between that vector and all columns
 of the other matrix are computed.
 
 `a` and `b` must have the same number of columns if neither of the two is a
-vector. `r` must be a vector of length `maximum(size(a, 2), size(b, 2))`.
+vector. `r` must be an array of length `maximum(size(a, 2), size(b, 2))`.
+
+!!! note
+    If both `a` and `b` are vectors, the generic, iterator-based method of
+    `colwise` applies.
 """
 function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix)
     require_one_based_indexing(r, a, b)
@@ -129,12 +137,12 @@ Compute distances between corresponding columns of `a` and `b` according to
 distance `metric`. Exactly one of `a` or `b` can be a vector, in which case the
 distance between that vector and all columns of the other matrix are computed.
 
+`a` and `b` must have the same number of columns if neither of the two is a
+vector.
+
 !!! note
     If both `a` and `b` are vectors, the generic, iterator-based method of
     `colwise` applies.
-
-`a` and `b` must have the same number of columns if neither of the two is a
-vector.
 """
 function colwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix)
     n = get_common_ncols(a, b)
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -152,11 +152,13 @@ function _pairwise!(r::AbstractMatrix, metric::SemiMetric, a)
     size(r) == (n, n) || throw(DimensionMismatch("Incorrect size of r."))
     itr = Iterators.product(enumerate(a), enumerate(a))
     @inbounds for ((i, ai), (j, aj)) in itr
-        r[i, j] = i > j ? 
-            metric(ai, aj) :
-            i == j ?
-                0 :
-                r[j,i]
+        r[i, j] = if i > j
+            metric(ai, aj)
+        elseif i == j
+            0
+        else
+            r[j, i]
+        end
     end
     r
 end
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -36,7 +36,7 @@ result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
 result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
 # Promote Arrays and Numbers to types
-result_type(dist, a, b) = result_type(dist, eltype(first(a)), eltype(first(b)))
+result_type(dist, a, b) = result_type(dist, typeof(first(a)), typeof(first(b)))
 result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
 result_type(dist, a::Number, b::Number) = result_type(dist, typeof(a), typeof(b))
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -26,15 +26,19 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
 # Generic functions
 
 """
-    result_type(dist::PreMetric, Ta::Type, Tb::Type) -> T
-    result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) -> T
+    result_type(dist, Ta::Type, Tb::Type) -> T
+    result_type(dist, a, b) -> T
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
-result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
+# Promote Arrays and Numbers to types
+result_type(dist, a, b) = result_type(dist, eltype(first(a)), eltype(first(b)))
+result_type(dist, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(dist, a::Number, b::Number) = result_type(dist, typeof(a), typeof(b))
 
 # Generic column-wise evaluation
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -36,7 +36,10 @@ result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
 result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
 # Promote Arrays and Numbers to types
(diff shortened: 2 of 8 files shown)


dkarrasch:
@mkborregaard You seemed to be interested in this.

src/generic.jl
```diff
@@ -26,15 +26,19 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
 """
-result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback
-result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) = result_type(dist, eltype(a), eltype(b))
+result_type(::PreMetric, ::Type, ::Type) = Float64 # fallback in Distances
+result_type(f, a::Type, b::Type) = typeof(f(oneunit(a), oneunit(b))) # don't require `PreMetric` subtyping
 
+# Promote Arrays and Numbers to types
+result_type(dist, a, b) = result_type(dist, typeof(first(a)), typeof(first(b)))
```
dkarrasch, line 39: Should we do some `Base.IteratorEltype`-magic, depending on whether it returns `HasEltype` or `EltypeUnknown`? It would be cool to avoid reading out the `first` elements, particularly for empty iterables.

johnnychen94: Hmmm, will it be problematic to also use `eltype` here? Otherwise, it seems to be clearer to use

```julia
result_type(dist, a, b) = result_type(dist, first(a), first(b))
```

johnnychen94: Sounds like a good idea.

```julia
result_type(dist, a, b) = _result_type(dist, _eltype(a), _eltype(b))
_eltype(a) = __eltype(Base.IteratorEltype(a), a)
__eltype(::Base.HasEltype, a) = eltype(a)
__eltype(::Base.EltypeUnknown, a) = typeof(first(a))
```

Something like this, maybe.

dkarrasch: Yeah, that's cool. That let's `hamming("", "")` work out of the box. Nondegenerate cases were no problem before.

test/test_dists.jl
```diff
@@ -130,7 +130,10 @@ end
     @test chebyshev(a, b) == 1.0
     @test braycurtis(a, b) === 1/3
     @test minkowski(a, b, 2) == 1.0
-    @test hamming(a, b) == 1
+    @test hamming(a, b) === 1
+    @test hamming("martha", "marhta") === 2
+    @test hamming("es an ", " vs an") === 6
+    @test hamming("", "") === 0
```
johnnychen94, line 136: This is interesting

johnnychen94 (approved):
Didn't look at every detail; it generally looks good to me. Approval of the goal.

Might need a benchmark to detect potential regression.

---

ImageDistances might get even simplified with this change.

dkarrasch:
I haven't generalized `*Mahalanobis`, because that looks so linear-algebra-y. Otherwise, I think, all metrics now work with abstract iterables, and are tested for `Iterators` and `Generators`. I'll try to do some benchmarks. In a few cases, I have left both a specific `AbstractVector` version and an unconstrained version. I'll try to benchmark whether there is any benefit from the constrained versions. Any comments or concerns are welcome, of course.

dkarrasch:
This is ready for a thorough review. Additional use cases and tests to be considered would be also welcome.

test/test_dists.jl
```diff
@@ -150,45 +153,50 @@ end
...
-
-            w = rand(Float64, size(x))
-            @test wsqeuclidean(x, y, w) ≈ dot((x - vec(y)).^2, w)
-            @test weuclidean(x, y, w) == sqrt(wsqeuclidean(x, y, w))
-            @test wcityblock(x, y, w) ≈ dot(abs.(x - vec(y)), w)
-            @test wminkowski(x, y, w, 2) ≈ weuclidean(x, y, w)
+            for (x, y) in ((x, y),
+                          ((Iterators.take(x, 4), Iterators.take(y, 4))),# iterator
+                          (((x[i] for i in 1:length(x)), (y[i] for i in 1:length(y)))),# generator
+                          )
```
nalimilan, line 162:
```suggestion
            for (x, y) in ((x, y),
                           ((Iterators.take(x, 4), Iterators.take(y, 4))), # iterator
                           (((x[i] for i in 1:length(x)), (y[i] for i in 1:length(y)))), # generator
                          )
```

test/test_dists.jl
```diff
@@ -228,22 +236,27 @@ end
...
-        @test renyi_divergence([0.0, 1.0, 0.0], [0.0, 0.5, 0.5], Inf) ≈ log(2.0)
-        @test renyi_divergence(p, q, 1) ≈ kl_divergence(p, q)
 
         pm = (p + q) / 2
-        jsv = kl_divergence(p, pm) / 2 + kl_divergence(q, pm) / 2
-        @test js_divergence(p, q) ≈ jsv
+        for (r, p, pm) in ((r, p, pm),
+                       (Iterators.take(r, length(r)), Iterators.take(p, length(p)), Iterators.take(pm, length(pm))),
+                       ((r[i] for i in 1:length(r)), (p[i] for i in 1:length(p)), (pm[i] for i in 1:length(pm)))
+                      )
```
nalimilan, line 244: Alignment is weird here.

src/bhattacharyya.jl
```diff
@@ -30,18 +30,36 @@ function bhattacharyya_coeff(a::AbstractVector{T}, b::AbstractVector{T}) where {
 
     sqab / sqrt(asum * bsum)
 end
+function bhattacharyya_coeff(a, b)
```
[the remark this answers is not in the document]
dkarrasch: I'm not sure, and was hoping for somebody to step in. The method above has a loop that can be `@simd`ed because it has a single simple iterator. The same optimization doesn't seem to work for loops à la `for (ai, bi) in zip(a, b)`. So it's a performance issue, which I should have commented on perhaps. For the time being, I'll rearrange that via multiple dispatch and reduce code duplication.

src/bhattacharyya.jl
```diff
@@ -30,18 +30,36 @@ function bhattacharyya_coeff(a::AbstractVector{T}, b::AbstractVector{T}) where {
...
+end
 
 # Faster pair- and column-wise versions TBD...
 
 
 # Bhattacharyya distance
-(::BhattacharyyaDist)(a::AbstractVector{T}, b::AbstractVector{T}) where {T <: Number} = -log(bhattacharyya_coeff(a, b))
-(::BhattacharyyaDist)(a::T, b::T) where {T <: Number} = throw("Bhattacharyya distance cannot be calculated for scalars")
+(::BhattacharyyaDist)(a, b) = -log(bhattacharyya_coeff(a, b))
+(::BhattacharyyaDist)(::Number, ::Number) = throw("Bhattacharyya distance cannot be calculated for scalars")
```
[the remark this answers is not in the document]
dkarrasch: I've never used this, so I only relaxed the existing method slightly. Any suggestion is welcome. We may remove this method altogether, and have a `length(a) == 1` check in the generic one?

src/bhattacharyya.jl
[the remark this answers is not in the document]
nalimilan: Ah you're right that `@simd` can only be used for `AbstractArray`. That's a classic issue for which I hope we could find a solution.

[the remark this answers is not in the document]
nalimilan: BTW, note that the `T <: Number` constraint could be lifted in the `AbstractArray` method.

src/haversine.jl
```diff
@@ -10,20 +10,17 @@ struct Haversine{T<:Real} <: Metric
     radius::T
 end
 
-const VecOrLengthTwoTuple{T} = Union{AbstractVector{T}, NTuple{2, T}}
-
-function (dist::Haversine)(x::VecOrLengthTwoTuple, y::VecOrLengthTwoTuple)
+function (dist::Haversine)(x, y)
     length(x) == length(y) == 2 || haversine_error()
 
-    @inbounds begin
```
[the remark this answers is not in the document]
dkarrasch: I didn't know whether it has an effect on destructuring at all. I thought it wouldn't because there is no indexing involved, just iteration.

src/metrics.jl
```diff
@@ -291,28 +320,27 @@ function _evaluate(dist::UnionMetrics, a::Number, b::Number, p)
...
 
 for M in (metrics..., weightedmetrics...)
-    @eval @inline (dist::$M)(a::AbstractArray, b::AbstractArray) = _evaluate(dist, a, b, parameters(dist))
-    if M != SpanNormDist
-        @eval @inline (dist::$M)(a::Number, b::Number) = _evaluate(dist, a, b, parameters(dist))
-    end
+    @eval @inline (dist::$M)(a, b) = _evaluate(dist, a, b, parameters(dist))
+    # if M != SpanNormDist
+    #     @eval @inline (dist::$M)(a::Number, b::Number) = _evaluate(dist, a, b, parameters(dist))
+    # end
```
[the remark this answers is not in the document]
dkarrasch: It has a special method, that seems to deviate from `_evaluate`(like less operations or something). Will remove, was just checking if everything works without it.

src/bhattacharyya.jl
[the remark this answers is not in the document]
nalimilan: Currently arrays with length 1 are allowed? I wonder why numbers wouldn't be allowed if we allow other iterators of length 1.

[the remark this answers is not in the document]
dkarrasch: At least I don't see any check. But I agree with you, what should be wrong about scalars if length-1 arrays work just fine.

src/haversine.jl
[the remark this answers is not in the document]
nalimilan: But iteration relies on indexing for arrays. It's worth checking at least, or keeping if in doubt.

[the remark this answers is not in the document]
dkarrasch: Does this confirm what you say?
```julia
julia> Meta.@lower x, y = rand(3)
:($(Expr(:thunk, CodeInfo(
    @ none within `top-level scope'
1 ─ %1 = rand(3)
│   %2 = Base.indexed_iterate(%1, 1)
│   %3 = Core.getfield(%2, 1)
│        x = %3
│        #s1 = Core.getfield(%2, 2)
│   %6 = Base.indexed_iterate(%1, 2, #s1)
│   %7 = Core.getfield(%6, 1)
│        y = %7
└──      return %1
))))
```
Seems so, right?

src/bhattacharyya.jl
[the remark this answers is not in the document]
dkarrasch: In the coming commit, I simply remove these methods, since they work via the iterator path, and return the same result as length-1 vectors.

src/metrics.jl
```diff
@@ -291,28 +320,27 @@ function _evaluate(dist::UnionMetrics, a::Number, b::Number, p)
...
-euclidean(a::AbstractArray, b::AbstractArray) = Euclidean()(a, b)
-euclidean(a::Number, b::Number) = Euclidean()(a, b)
+euclidean(a, b) = Euclidean()(a, b)
 
 # Weighted Euclidean
 @inline eval_op(::WeightedEuclidean, ai, bi, wi) = abs2(ai - bi) * wi
 eval_end(::WeightedEuclidean, s) = sqrt(s)
-weuclidean(a::AbstractArray, b::AbstractArray, w::AbstractArray) = WeightedEuclidean(w)(a, b)
-weuclidean(a::Number, b::Number, w::Real) = WeightedEuclidean([w])(a, b)
+weuclidean(a, b, w::AbstractArray) = WeightedEuclidean(w)(a, b)
```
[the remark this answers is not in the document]
dkarrasch: I completely agree. I wanted to focus on the `a`,`b` arguments first. In the spirit of this PR, we could basically remove any constraint on the weights/periods parameter, and leave it to the user to provide something iterable. All existing code would work as before.

src/metrics.jl
```diff
@@ -397,32 +420,31 @@ function eval_end(::CosineDist, s)
...
     max(1 - ab / (sqrt(a2) * sqrt(b2)), 0)
 end
-cosine_dist(a::AbstractArray, b::AbstractArray) = CosineDist()(a, b)
-cosine_dist(a::Number, b::Number) = CosineDist()(a, b)
+cosine_dist(a, b) = CosineDist()(a, b)
 
 # CorrDist
-_centralize(x::AbstractArray) = x .- mean(x)
-(dist::CorrDist)(a::AbstractArray, b::AbstractArray) = CosineDist()(_centralize(a), _centralize(b))
+_centralize(x) = x .- mean(x)
```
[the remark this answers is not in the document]
dkarrasch: No, it seems to work for abstract iterables:
```julia
julia> itr = (i for i in 1:3)
Base.Generator{UnitRange{Int64},typeof(identity)}(identity, 1:3)

julia> itr .- mean(itr)
3-element Array{Float64,1}:
 -1.0
  0.0
  1.0
```
And I hope it's tested for such, actually. I'll check.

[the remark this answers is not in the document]
dkarrasch: Yes, it's tested for iterators and generators.

src/bhattacharyya.jl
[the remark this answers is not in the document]
dkarrasch: I refactored the `bhattacharyya` code and use multiple dispatch to distinguish the simd vs non-simd cases.

src/metrics.jl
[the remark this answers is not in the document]
dkarrasch: This is now all relaxed! The types are now of the form
```julia
struct Weighted*{W} <: [SemiMetric]
    weights::W
end
```
and it's left to the user to provide meaningful `weights`. The requirement is basically the same as for `a` and `b`: something iterable with known length.

src/metrics.jl
```diff
@@ -220,13 +220,42 @@ result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, ::Nothing) where {Ta,Tb}
...
 
-Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b::AbstractArray)
+Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a, b)
     _evaluate(d, a, b, parameters(d))
 end
-_evaluate(dist::UnionMetrics, a::Number, b::Number) = _evaluate(dist, a, b, parameters(dist))
 
 # breaks the implementation into eval_start, eval_op, eval_reduce and eval_end
 
+Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a, b, ::Nothing)
```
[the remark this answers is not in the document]
dkarrasch: I have tried to that in ef6c646. That commit works, and it seems to distribute the exact same code into inlined helper functions, and yet it introduces patterns like this in `@code_typed euclidean(rand(100), rand(100))`:
```
%119 = Base.add_int(%100, 1)::Int64
│            $(Expr(:loopinfo, Symbol("julia.simdloop"), nothing))::Any
└────        goto #37
45 ┄─ %122 = φ (#37 => %99, #35 => 0.0)::Float64
46 ┄─ %123 = φ (#45 => %122, #34 => 0.0)::Float64
│     %124 = Base.lt_float(%123, 0.0)::Bool
└────        goto #48 if not %124
47 ──        invoke Base.Math.throw_complex_domainerror(:sqrt::Symbol, %123::Float64)::Union{}
└────        $(Expr(:unreachable))::Union{}
48 ┄─ %128 = Base.Math.sqrt_llvm(%123)::Float64
└────        goto #49
49 ──        goto #50
50 ──        goto #98
51 ──        Base.Math.sqrt_llvm(0.0)::Float64
│     %133 = Base.bitcast(UInt64, 1)::UInt64
│     %134 = Base.bitcast(UInt64, 1)::UInt64
│     %135 = Base.sub_int(%133, %134)::UInt64
│     %136 = Base.arraylen(a)::Int64
│     %137 = Base.sle_int(0, %136)::Bool
│     %138 = Base.bitcast(UInt64, %136)::UInt64
│     %139 = Base.ult_int(%135, %138)::Bool
```
I'm not very good at reading this, but it seems like after each loop iteration, it is checking whether it should apply `sqrt`. Also, the `Base.bitcast` calls all completely absent in the current master version. I hope I have done stupid, easily fixable mistake, because I kind of like the design in that commit.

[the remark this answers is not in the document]
dkarrasch: I believe one can measure these extra operations in `@btime`, though the slow-down is not dramatic.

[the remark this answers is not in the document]
nalimilan: To be clear, you're comparing the current:
```julia
            @inbounds for (Ia, Ib) in zip(eachindex(a), eachindex(b))
                ai = a[Ia]
                bi = b[Ib]
                s = eval_reduce(d, s, eval_op(d, ai, bi))
            end
```
with the new:
```julia

        @inbounds for (ai, bi) in zip(a, b)
            s = eval_reduce(d, s, eval_op(d, ai, bi))
        end
```

right?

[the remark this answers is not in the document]
dkarrasch: No, I'm comparing the `@simd`ed loops paths. You may need to turn off the diff show to see better what the path is in the last commit.

src/generic.jl
```diff
@@ -104,18 +146,34 @@ function _pairwise!(r::AbstractMatrix, metric::PreMetric,
...
+    require_one_based_indexing(r)
+    n = length(a)
+    size(r) == (n, n) || throw(DimensionMismatch("Incorrect size of r."))
+    itr = Iterators.product(enumerate(a), enumerate(a))
+    @inbounds for ((i, ai), (j, aj)) in itr
+        r[i, j] = i > j ? 
+            metric(ai, aj) :
+            i == j ?
+                0 :
+                r[j,i]
```
nalimilan, line 159: I find this a bit hard to read. How about something like this (many variants are possible)?
```suggestion
        r[i, j] = if i > j
            metric(ai, aj)
        elseif
            0        
        else
            r[j, i]
        end
```

src/generic.jl
```diff
@@ -32,39 +32,58 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
 result_type(dist, a, b) = result_type(dist, _eltype(a), _eltype(b))
+# description of approach:
+# (a) for generic iterators, rely on Base.IteratorEltype(a)
 _eltype(a) = __eltype(Base.IteratorEltype(a), a)
-__eltype(::Base.HasEltype, a) = eltype(a)
-__eltype(::Base.EltypeUnknown, a) = typeof(first(a))
+__eltype(::Base.HasEltype, a) = _eltype(eltype(a))
+__eltype(::Base.EltypeUnknown, a) = _eltype(typeof(first(a)))
+# (b) for arrays (of arrays), go down the eltype route, until you hit a non-array eltype and return that
+_eltype(a::AbstractArray) = _eltype(eltype(a))
```
nalimilan, line 45: Is it really needed to special-case arrays? The generic code for iterators should work AFAICT.

Also, do you think arrays of arrays of arrays will work? If we want to allow that, maybe there should be a test?

src/metrics.jl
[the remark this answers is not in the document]
nalimilan: OK. I'm not completely sure how to read that either, but `goto #37` seems to indicate that the core of the loop won't reach the `sqrt` call: that will only happen at the end. `@code_llvm` doesn't show any `sqrt` call in the SIMD loop either.

But if you say you can see the difference when benchmarking, maybe reintroduce the old loop. Given that we need to keep the check for the `@simd` path, the deduplication is more limited than I hoped anyway.

src/generic.jl
dkarrasch: Yes, without that specialization, you only ask for the `eltype` once, and in the case of, say, vectors of vectors you get some array type. Then, `result_type` tries to construct a `oneunit` of that array type... and fails. This is consistent with the explicit `eltype(eltype(a))` construction in #188.

src/metrics.jl
```diff
@@ -263,56 +241,85 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
...
     end
 end
 
-_evaluate(dist::UnionMetrics, a::Number, b::Number, ::Nothing) = eval_end(dist, eval_op(dist, a, b))
-function _evaluate(dist::UnionMetrics, a::Number, b::Number, p)
+evaluate(dist::UnionMetrics, a::Number, b::Number, ::Nothing) = eval_end(dist, eval_op(dist, a, b))
+function evaluate(dist::UnionMetrics, a::Number, b::Number, p)
     length(p) != 1 && throw(DimensionMismatch("inputs are scalars but parameters have length $(length(p))."))
     eval_end(dist, eval_op(dist, a, b, first(p)))
 end
```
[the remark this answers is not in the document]
dkarrasch: I'm not sure. We save the creation of a zero in `eval_start` and a call to `eval_reduce` (often `0 + eval_op(d, a, b)` in the first step). I'm not sure if the compiler is able to compile that away.

src/generic.jl
dkarrasch: And we do have tests for the array of array case, basically copied from #188.

dkarrasch:
I reverted the `UnionMetric` changes, the code duplication (except for size checks) is not very big after all. I don't see any regression in the classic vector case (and I'm not so sure anymore the regression I observed earlier was real). This supersedes now #192 and #188, I included those tests (with mild modifications) to make sure the relaxation of `result_type` and the generalization to arrays of arrays and iterable objects is consistent. The `result_type` relaxation makes it now obsolete to define the `Float64` fallback. Except for empty iterables with unknown eltype, `result_type` is set up to infer the result type for arrays (of arrays) and iterables alike.

dkarrasch:
I also included updated docstrings, which should help building a Documenter documentation at some point.

I got a little obsessive about the "lazy evaluation" approach. If you look closer, then `colwise` is nothing but `map(metric, zip(eachcol(A), eachcol(B))` (and similarly for iterators), and `pairwise(..., dims=2)` is nothing but `map(Iterators.product(eachcol(A), eachcol(B))` (or `eachrow` for `dims=1` and similarly for iterators). Replacing `map` by `Iterators.map` makes the whole construct lazy, i.e., a generator. Does anybody see any value in exposing this lazy construct, which could be, by default, eagerly evaluated via `collect`? One naive use case I can imagine is computing k-nearest neighbors within the data set via brute force. That shouldn't require having the total distance matrix in memory?

src/generic.jl
```diff
@@ -89,16 +109,32 @@ function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::Abs
     r
 end
 
-function colwise!(r::AbstractArray, metric::SemiMetric, a::AbstractMatrix, b::AbstractVector)
-    colwise!(r, metric, b, a)
-end
+"""
+    colwise(r::AbstractMatrix, metric::PreMetric, a, b)
```
nalimilan, line 113:
```suggestion
    colwise(metric::PreMetric, a, b)
```

src/generic.jl
```diff
@@ -89,16 +109,32 @@ function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::Abs
...
+
+Compute distances between corresponding elements of the iterable collections
+`a` and `b` according to distance `metric`.
 
+`a` and `b` must have the same number of elements, `r` must be a vector of length
+`length(a) == length(b)`.
```
nalimilan, line 119:
```suggestion
`a` and `b` must have the same number of elements (`length(a) == length(b)`).
```

src/generic.jl
```diff
@@ -89,16 +109,32 @@ function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::Abs
...
+"""
 function colwise(metric::PreMetric, a, b)
     n = get_common_length(a, b)
     r = Vector{result_type(metric, a, b)}(undef, n)
     colwise!(r, metric, a, b)
 end
 
+"""
+    colwise(r::AbstractMatrix, metric::PreMetric,
+              a::AbstractMatrix, b::AbstractMatrix)
```
nalimilan, line 129:
```suggestion
    colwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix)
```

src/generic.jl
```diff
@@ -89,16 +109,32 @@ function colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::Abs
...
+
+Compute distances between each corresponding columns of `a` and `b` according
+to distance `metric`. Exactly one of `a` or `b` can be a vector, in which case
+the distance between that vector and all columns of the other matrix are computed.
+
+`a` and `b` must have the same number of columns if neither of the two is a
+vector. `r` must be a vector of length `maximum(size(a, 2), size(b, 2))`.
```
nalimilan, line 136: `a` and `b` can't be vectors with this signature. Also `r` isn't used by this method.

src/generic.jl
```diff
@@ -245,8 +281,22 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
     end
 end
 
+"""
+    pairwise!(r::AbstractMatrix, metric::PreMetric, a, b)
+    pairwise!(r::AbstractMatrix, metric::PreMetric, a)
```
nalimilan, line 286:
```suggestion
    pairwise!(r::AbstractMatrix, metric::PreMetric, a[, b])
```

src/generic.jl
```diff
@@ -245,8 +281,22 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
...
+If a single iterable `a` is provided, compute distances between its elements.
+
+`r` must be a matrix with size `length(a) × length(b)`.
+"""
+pairwise!(r::AbstractMatrix, metric::PreMetric, a, b) = _pairwise!(r, metric, a, b)
+pairwise!(r::AbstractMatrix, metric::PreMetric, a) = _pairwise!(r, metric, a)
+
 """
     pairwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix=a; dims)
+    pairwise(metric::PreMetric, a::AbstractMatrix; dims)
```
nalimilan, line 299: This is redundant with `b=a` above.
```suggestion
```

src/generic.jl
```diff
@@ -280,6 +323,21 @@ function pairwise(metric::PreMetric, a::AbstractMatrix;
     pairwise!(r, metric, a, dims=dims)
 end
 
+"""
+    pairwise(metric::PreMetric, a, b)
+    pairwise(metric::PreMetric, a)
```
nalimilan, line 328:
```suggestion
    pairwise(metric::PreMetric, a[, b])
```

src/metrics.jl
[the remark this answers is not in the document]
nalimilan: OK let's keep them then.

src/generic.jl
nalimilan: What I mean is that for arrays, `_eltype(a) = __eltype(Base.IteratorEltype(a), a)` should end up calling `__eltype(::Base.HasEltype, a) = _eltype(eltype(a))`, which is exactly the same as this definition. Am I missing something?

src/generic.jl
```diff
@@ -26,49 +26,115 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
+    n = get_common_length(a, b)
+    r = Vector{result_type(metric, a, b)}(undef, n)
+    colwise!(r, metric, a, b)
 end
 
+"""
+    colwise(r::AbstractMatrix, metric::PreMetric,
+              a::AbstractMatrix, b::AbstractMatrix)
+
+Compute distances between each corresponding columns of `a` and `b` according
```
nalimilan, line 131: Something to note is that these methods are inconsistent with the ones treating `a` and `b` as iterators of columns: a matrix of vectors will be treated differently from a vector of the same vectors. That's probably OK in practice, but that's one of the reasons why I'd like to move to requiring explicitly writing `pairwise(d, eachol(a), eachcol(b))` in the longer term. That way we won't need `dims` anymore.

src/generic.jl
```diff
@@ -185,8 +281,22 @@ function pairwise!(r::AbstractMatrix, metric::PreMetric, a::AbstractMatrix;
...
+"""
+pairwise!(r::AbstractMatrix, metric::PreMetric, a, b) = _pairwise!(r, metric, a, b)
+pairwise!(r::AbstractMatrix, metric::PreMetric, a) = _pairwise!(r, metric, a)
+
 """
     pairwise(metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix=a; dims)
+    pairwise(metric::PreMetric, a::AbstractMatrix; dims)
 
 Compute distances between each pair of rows (if `dims=1`) or columns (if `dims=2`)
 in `a` and `b` according to distance `metric`. If a single matrix `a` is provided,
```
nalimilan, line 302: Something to note is that these methods are inconsistent with the ones treating `a` and `b` as iterators of columns: a matrix of vectors will be treated differently from a vector of the same vectors. That's probably OK in practice, but that's one of the reasons why I'd like to move to requiring explicitly writing `pairwise(d, eachol(a), eachcol(b))` in the longer term. That way we won't need `dims` anymore.

src/generic.jl
```diff
@@ -26,49 +26,115 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
 # Generic functions
 
 """
-    result_type(dist::PreMetric, Ta::Type, Tb::Type) -> T
-    result_type(dist::PreMetric, a::AbstractArray, b::AbstractArray) -> T
+    result_type(dist, Ta::Type, Tb::Type) -> T
+    result_type(dist, a, b) -> T
 
 Infer the result type of metric `dist` with input type `Ta` and `Tb`, or input
 data `a` and `b`.
```
nalimilan, line 33:
```suggestion
Infer the result type of metric `dist` with input types `Ta` and `Tb`, or element types
of iterators `a` and `b`.
```

nalimilan:
> I got a little obsessive about the "lazy evaluation" approach. If you look closer, then `colwise` is nothing but `map(metric, zip(eachcol(A), eachcol(B))` (and similarly for iterators), and `pairwise(..., dims=2)` is nothing but `map(Iterators.product(eachcol(A), eachcol(B))` (or `eachrow` for `dims=1` and similarly for iterators). Replacing `map` by `Iterators.map` makes the whole construct lazy, i.e., a generator. Does anybody see any value in exposing this lazy construct, which could be, by default, eagerly evaluated via `collect`? One naive use case I can imagine is computing k-nearest neighbors within the data set via brute force. That shouldn't require having the total distance matrix in memory?

Given that using `Iterators.map` manually is relatively easy, is it really worth providing something else? I guess we could add an argument to `pairwise` (say, `lazy`), but that can be done at any time if somebody requests it.

src/generic.jl
dkarrasch: I see. I removed the special casing now, assuming that the "bottom type" is reached when `eltype(T) === T`. Otherwise, both arrays (of arrays/iterators) and iterators (of arrays/iterators) should go down the route until you hit that condition.

src/generic.jl
dkarrasch: You're right, not with this signature. But we do allow other signatures as well, I'll list them above.

src/generic.jl
dkarrasch: I don't like this `b=a` thing, because this is not how we do it internally, but that doesn't matter to the user I guess, so it's fine for the docstring.

src/generic.jl
dkarrasch: That would be very nice, because it redirects the matrix-based method to the iterator-based method, and one could get rid of the matrix-based ones. The only issue I see is that for the specialized `*Euclidean` (and a few others) distances, where we do need the underlying matrix for performance reason, I don't seem to be able to unwrap it from the `eachcol` generator.

src/generic.jl
dkarrasch: I'm not exactly sure I understand the inconsistency, actually. Could you please sketch an application case?

dkarrasch:
Alright, I better stop polishing now. This now fixes a whole bunch of issues and/or adds new features. I believe this is in good shape now. Comments and reviews are welcome. Shall we bump the patch number, or the minor version? It's rather difficult to propagate "next level versions" through the ecosystem, and since this doesn't break any existing functionality, maybe a patch is enough?

src/generic.jl
nalimilan: Fortunately https://github.com/JuliaLang/julia/pull/32310 should allow us to retrieve the underlying matrix!

dkarrasch: Yes, I looked at that one a bit today. Will they let it go into v1.6? I wonder how the ecosystem is going to adapt to v1.6 being the new LTS, and how fast packages will really drop 1.6- support. In many cases, there is no hard reason, only soft ones.

src/generic.jl
nalimilan: For example, `pairwise(d, [a, b, c, d])` vs. `pairwise(d, reshape([a, b, c, d], 2, 2))` with `a`, `b`, `c` and `d` vectors of numbers.

test/test_dists.jl
```diff
@@ -717,6 +717,14 @@ end
     @test bregman(F, ∇, p, q) ≈ ISdist(p, q)
 end
 
+@testset "CartesianIndex" begin
+    A = reshape(collect(1:9), 3, 3)
+    inds1 = findall(iseven, A)
+    inds2 = findall(isodd, A)
+    @test sum(pairwise(SqEuclidean(), inds1, inds2)) == 52
+    @test euclidean(inds1[1], inds1[1]) === 0.0
+end
```
nalimilan: Does this cover a real use case? I find it surprising that somebody would want to do that.

src/generic.jl
```diff
@@ -26,49 +26,116 @@ evaluate(dist::PreMetric, a, b) = dist(a, b)
...
-        r[j] = metric(view(a, :, j), b)
+    @inbounds for (rj, aj) in enumerate(axes(a, 2))
+        r[rj] = metric(view(a, :, aj), b)
     end
     r
 end
 
+"""
+    colwise!(r::AbstractArray, metric::PreMetric,
+             a::AbstractMatrix, b::AbstractMatrix)
```
nalimilan, line 89: Need to show signatures with vectors here too.

nalimilan (commented):
Looks good. Just two small things.

test/test_dists.jl
dkarrasch: There was a feature request in #177. But we could remove it for now and clarify there what the actual intention and the use are.

src/generic.jl
dkarrasch: Ok, then I understood you correctly. Out of the two calls you mentioned, only the first one works. The second one fails because it treats `a`, `b`, `c` and `d` like numbers, but then calls like `abs(a)` (or whatever necessary) fail.

nalimilan: Yes. But if we were fully consistent, the second call would be equivalent to the first one, since matrices are just one kind of iterator.

test/test_dists.jl
nalimilan: Yes, I'd leave it out until we have a clearer view of why that would be useful. The fact that `CartesianIndex` isn't iterable seems to indicate that it's not intended to be used that way, and it would be absurd to have packages work around that everywhere...

nalimilan (approved):
Looks good if you leave `CartesianIndex` out for now. Thanks!

dkarrasch:
I'd merge and release today. Any opinions on how to bump the version number?

dkarrasch:
I merged the PR as is. We may include some others and then decide about the version bump.

Datseris:
@dkarrasch thanks you are a legend. With 1 PR closed like 10 issues!

dkarrasch:
Thanks, @Datseris. My next projects are world peace and healing cancer... in one PR! 🤣

nalimilan:
Thanks!  Regarding the release number, I think we have two options:
- tag 0.10.1 to avoid forcing dependencies to update their version requirements (since this isn't breaking)
- tag 1.0.0 so that we can more easily bump either the minor or patch version as appropriate in the future

johnnychen94:
This PR gives so many possibilities and thus a new start, so I'm voting for 1.0.0 😆

dkarrasch:
How about tagging 0.10.1, then have a v0.11 that deprecates the dims keyword and redirects to `eachcol`/`eachrow`, and then have a v0.12/v1 for a version that only contains iterator-based methods, except for array-based specializations for improved performance?

nalimilan:
Yes, if we anticipate breaking changes soon, better not tag 1.0. Maybe file an issue so that we can discuss these changes?

PR #196, opened 2020-12-21
Minor improvements, fixes, and code style

dkarrasch:
* fix order of matrix filling
* remove redundant matvec-mul in `Mahalanobis`
* remove unused argument names

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -802,7 +802,7 @@ end
 
 # CosineDist
 
-function _pairwise!(r::AbstractMatrix, dist::CosineDist,
+function _pairwise!(r::AbstractMatrix, ::CosineDist,
                     a::AbstractMatrix, b::AbstractMatrix)
     m, na, nb = get_pairwise_dims(r, a, b)
     mul!(r, a', b)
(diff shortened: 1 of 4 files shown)


src/metrics.jl
```diff
@@ -802,7 +802,7 @@ end
 
 # CosineDist
 
-function _pairwise!(r::AbstractMatrix, dist::CosineDist,
+function _pairwise!(r::AbstractMatrix, ::CosineDist,
```
nalimilan, line 805: Do we use the same style elsewhere?

dkarrasch: Yes, pretty much everywhere, like in the `eval_*` section of the same file, whenever we don't need `dist` explicitly, but only dispatch on the type in that argument slot.

src/metrics.jl
```diff
@@ -838,16 +838,16 @@ end
 #    of `_centralize` -- ~4x speed up
 _centralize_colwise(x::AbstractVector) = x .- mean(x)
 _centralize_colwise(x::AbstractMatrix) = x .- mean(x, dims=1)
-function colwise!(r::AbstractVector, dist::CorrDist, a::AbstractMatrix, b::AbstractMatrix)
+function colwise!(r::AbstractArray, ::CorrDist, a::AbstractMatrix, b::AbstractMatrix)
```
nalimilan, line 841: Sorry I had missed the `AbstractArray` change. I think it should be tested, otherwise it may regress without being noticed.

dkarrasch: I only realized that because you corrected my newly added docstring recently. 😉 We have that signature in the generic `colwise!` methods.

nalimilan (approved):

dkarrasch:
Is this good to go?

Issue #138, opened 2019-06-18
colwise function


musm:
Why is there only a  `colwise` function and not also a  matching rowwise function?
Personally I find `pairwise` better to use, so I'd also support deprecating colwise.

dkarrasch:
But `colwise` and `pairwise` have very different meanings here, independent of whether data points are viewed as columns or rows. `colwise` is meant to compute distances only between corresponding data points/columns. `pairwise` is meant to compute all pairwise distances, and can be used to compute the distances between matrices with different number of data points (say, columns), which is impossible for `colwise`, except for when one of the two data sets is a single data point, i.e., a vector.

musm:
Good point regarding pairwise (comment edited). I think I was getting at a `rowwise` that computes the distances between rows of the matrixes i.e. `rowwise(dist, X,Y) = colwise(dist, X', Y')

johnnychen94:
I suppose one of reason here is because julia matrix is stored in column-major order, so a row-wise loop might raise performance issue.

```julia
julia> x = [1 2 3; 4 5 6; 7 8 9]
3×3 Array{Int64,2}:
 1  2  3
 4  5  6
 7  8  9

julia> x[:]
9-element Array{Int64,1}:
 1
 4
 7
 2
 5
 8
 3
 6
 9
```

nalimilan:
With https://github.com/JuliaLang/julia/pull/32310 we should probably drop `colwise(dist, x, y)` in favor of `map(d, eachol(x), eachcol(y))`. Cc: @simonbyrne

dkarrasch:
I guess we should keep `colwise(dist, x, y)`, make `map(d, eachcol(x), eachcol(y))` the default, but allow specialized methods to optimize for performance.
```julia
using Distances, BenchmarkTools
d = Euclidean(); a = rand(5, 100); b = rand(5, 100);
@btime map($d, $(eachcol(a)), $(eachcol(b))); # 3.090 μs (304 allocations: 13.45 KiB)
@btime colwise($d, $a, $b); # 380.907 ns (1 allocation: 896 bytes)
```

simonbyrne:
@dkarrasch the point of https://github.com/JuliaLang/julia/pull/32310 is that you could write specialized versions that can leverage the memory layout: in this case, you could do:
```julia
Base.map(d::Distance, a::EachCol, b::EachCol) = colwise(d, parent(a), parent(b))
```

mroavi:
[...]

PR #197, opened 2020-12-21
Draft for deprecation route

dkarrasch:
I went a ahead and started a draft for how one could proceed further, making the package more/solely iterator-based. A few thoughts and suggestions:

* replace `rowwise` by `zipwise` and ask for `eachcol` or `eachrow` iterators, thus removing the necessity to even think about things like a `rowwise` command (cf. #138);
* essentially remove the `dims` keyword in `pairwise`; this is currently only half way, because some specializations do require an underlying matrix, but assuming we are typically given row- or column-iterators, we don't have access to that matrix currently (for the full way version, this requires JuliaLang/julia#32310).

This is breaking, and could form one piece of a transitional version, bumped in the minor version. In a next step, we could remove the matrix-based legacy functions, except those needed for performance reasons, which requires (a) JuliaLang/julia#32310 and (b) a lower Julia bound v1.6/1.7(?), depending on where this feature is included.

### DIFF
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -44,106 +44,15 @@ __eltype(::Base.EltypeUnknown, a) = _eltype(typeof(first(a)))
 
 # Generic column-wise evaluation
 
-@deprecate colwise!(r::AbstractArray, metric::PreMetric, a, b) zipwise!(r, metric, a, b)
+@deprecate colwise!(r::AbstractArray, metric::PreMetric, a::AbstractVector, b::AbstractMatrix) broadcast!(metric, r, Ref(a), eachcol(b))
+@deprecate colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::AbstractVector) broadcast!(metric, r, eachcol(a), Ref(b))
+@deprecate colwise!(r::AbstractArray, metric::PreMetric, a::AbstractMatrix, b::AbstractMatrix) broadcast!(metric, r, eachcol(a), eachcol(b))
+@deprecate colwise!(r::AbstractArray, metric::PreMetric, a, b) broadcast!(metric, r, a, b)
 
-"""
-    zipwise!(r::AbstractArray, metric::PreMetric, a, b)
-
-Compute distances between corresponding elements of the iterable collections
-`a` and `b` according to distance `metric`, and store the result in `r`. Exactly
-one of `a` or `b` can be a length-1 iterator, in which case the distance between
-that object and all objects of the respective other iterator are computed.
-
-`r` must be an array of length `maximum(length(a), length(b))`.
-"""
-function zipwise!(r::AbstractArray, metric::PreMetric, a, b)
-    na = length(a)
-    nb = length(b)
-    if na == nb
-        length(r) == na || throw(DimensionMismatch("Incorrect size of r."))
-        @inbounds for (j, ab) in enumerate(zip(a, b))
-            r[j] = metric(ab...)
-        end
-    elseif na == 1
-        length(r) == nb || throw(DimensionMismatch("Incorrect size of r."))
-        @inbounds for (j, ab) in enumerate(zip(Iterators.repeated(a, nb), b))
-            r[j] = metric(ab...)
-        end
-    elseif nb == 1
-        length(r) == nb || throw(DimensionMismatch("Incorrect size of r."))
-        @inbounds for (j, ab) in enumerate(zip(a, Iterators.repeated(b, na)))
-            r[j] = metric(ab...)
-        end
-    else
-        throw(DimensionMismatch("The lengths of a and b must match."))
-    end
-    return r
-end
-
-# legacy methods, would be better to enforce iterators more strongly (colwise vs rowwise)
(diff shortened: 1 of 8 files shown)


dkarrasch:
I just found/recalled the proposal to remove `rowwise` altogether in favor of `map(dist, each[col/row](a), each[col/row](b))`. While this works just fine, unfortunately the corresponding in-place `map!` doesn't, because it seems to require that the arguments are arrays rather than just iterators. Something to keep in mind.

src/generic.jl
```diff
@@ -44,94 +44,85 @@ __eltype(::Base.EltypeUnknown, a) = _eltype(typeof(first(a)))
 
 # Generic column-wise evaluation
 
+@deprecate colwise!(r::AbstractArray, metric::PreMetric, a, b) zipwise!(r, metric, a, b)
+
 """
-    colwise!(r::AbstractArray, metric::PreMetric, a, b)
+    zipwise!(r::AbstractArray, metric::PreMetric, a, b)
```
nalimilan, line 50: TBH I don't find `zipwise` to be a super natural name. This exists only because `map!` only accepts arrays? We could define `map!` methods that work on any iterators even if the fallback `map!` method doesn't.

dkarrasch: Yea, this was just a proposal, since colwise wouldn't make much sense anymore. I realized the whole `map!` issue only after I did the "zipwise" commit. I guess `map!` doesn't exist for general iterators because they may not have a known length, but one could indeed work around that.

nalimilan: I can't see a good reason for `map!` not to support arbitrary iterators as inputs given that `map` supports them. The docstring mentions "collection" without any details. We could file an issue in Julia to discuss this.

nalimilan:
I'm afraid we can't deprecate `pairwise` on matrices until we drop support for all Julia versions which don't include JuliaLang/julia#32310. And the only chance this happens before several years is that JuliaLang/julia#32310 gets merged before 1.7 (probable), 1.7 becomes the LTS (less probable), and we drop support for versions before it (totally possible). However, even before that we could mention in the docs that the preferred form is to use `eachcol`/`eachrow` when using recent Julia versions.

src/generic.jl
dkarrasch: https://github.com/JuliaLang/julia/issues/38344

dkarrasch:
I agree. Without JuliaLang/julia#32310 and `map!` for iterators, are we going to deprecate anything at all or just leave things basically as they are? I'm fine to close this, or just have a recommendation for using `eachcol` and `eachrow`, though it's less clear why we would recommend away from something that works and continues to do so.

nalimilan:
Looks like we're blocked waiting for changes in Julia. Though we could imagine adding `map!` methods for `PreMetric` right now if we wanted to.

dkarrasch:
Here is an interesting thing. I just gave the "custom" `map!` approach a go, and realized that there is the issue with computing `colwise` distances between a vector and a matrix, a matrix and a vector, and a matrix and a matrix. So we need to be able to "blow up" vectors to the correct size of the other matrix. But hey, this is what broadcasting already knows how to do. So I guess, instead of `colwise!(r, metric, a, b)`, we should write `r .= metric.(a, b)`, or `Ref(a)` or `Ref(b)`, depending on which of the two is a vector. I guess people usually know which of the two is a vector, and would know which one to wrap? For non-mutating colwise, you would just write `r = metric.(a, b)`. 😄  Referring to broadcasting would relieve us from doing the dirty `map!` hack for iterators.

dkarrasch:
Hm, I forgot about the specializations of `colwise`... will think about it tomorrow.

src/mahalanobis.jl
```diff
@@ -32,21 +32,6 @@ end
...
-end
-
-function colwise!(r::AbstractArray, dist::SqMahalanobis, a::AbstractVector, b::AbstractMatrix)
-    Q = dist.qmat
-    m, n = get_colwise_dims(size(Q, 1), r, a, b)
-    z = a .- b
-    Qz = Q * z
-    dot_percol!(r, Q * z, z)
-end
-
```
nalimilan, line 49: Why do you say these are "useless"? Shouldn't we make them specializations of `broadcast!` (and `map!` probably too) once JuliaLang/julia#32310 is merged?

dkarrasch: Yes, sorry, I was too quick with those ones. The Mahalanobis methods are the only useful ones, and, as you say, are another use case for the iterator type. But without lower bounding the package to a Julia version that includes that PR, we can't deprecate colwise for the exact same reason as with `pairwise`.

dkarrasch:
I'm not sure this is the right time to go in this direction. We can deprecate neither `colwise` altogether nor pairwise on matrices, i.e., the user-facing API, since we need them for important specializations. We could rewrite and simplify quite some code internally using `eachcol` and `eachrow`, but that forces us to lift the lower julia bound to v1.1, which is neither LTS nor the current version. I think we should revisit this once JuliaLang/julia#32310 is merged and then we'll know in which environment we're working.

PR #198, opened 2020-12-24
add SphericalAngle distance

### LINKED ISSUE #195: Spherical angle distance
I often work with points on a sphere and need to compute distances between them. In principle, this should be covered by the `Haversine` distance in `Distances.jl`.
However, it performs multiple conversions on the way, which (slightly) affect performance and introduce extra consfusion. Namely, `Haversine` first converts inputs from degrees to radians, and then converts output from radians to linear distance units by multiplying by the radius.
So, I implemented a simple modification of that: `SphericalAngle` distance where all values are in the same units, in radians. I've been using this for quite some time already and put into my utility package. The code is available at https://gist.github.com/aplavin/6d6df09d9f17d19c70fd91ee07045a51. I think it is useful in any cases when points represent directions, do not lie on any specific sphere, and one works in natural units (radians).
If you think it can be included into `Distances.jl`, I can make a PR.

aplavin:
Closes #195.

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,6 +32,7 @@ This package also provides optimized functions to compute column-wise and pairwi
 * Bhattacharyya distance
 * Hellinger distance
 * Haversine distance
+* Spherical angle distance
 * Mean absolute deviation
 * Mean squared deviation
 * Root mean squared deviation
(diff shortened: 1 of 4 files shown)


dkarrasch:
This LGTM. I think we should mention the new distance in the README, and then this is ready to go.

aplavin:
Added to readme

aplavin:
Any plans to register a new version soon?..

PR #202, opened 2021-01-11
Clean up README

### LINKED ISSUE #204: New release?
I am very excited to see that there is now a general implementation for iterators, that will be very useful for some of my packages. However, it seems it is not released yet. Would it be possible to make a new release?

### LINKED ISSUE #206: Colwise and pairwise can return incorrect distances
Due to improper use of `@inbounds`, the `colwise` function can nondeterministically return incorrect results when used on arrays with offset axes due to out-of-bounds memory accesses.

```julia
julia> using Distances, OffsetArrays

julia> test_colwise() = sum(
               colwise(
                       Euclidean(),
                       OffsetArray(ones(11, 11), -5:5, -5:5),     # matrix of 1.0
                       OffsetArray(2 .* ones(11, 11), -5:5, -5:5) # matrix of 2.0
               )
       )
test_colwise (generic function with 1 method)

julia> [test_colwise() for _ in 1:10]
10-element Array{Float64,1}:
  Inf
  Inf
  Inf
 NaN
  32.49537442426318
  32.49537442426318
  32.49537442426318
  32.49537442426318
  32.49537442426318
  32.49537442426318
```

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L61

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L71

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L81

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L81

It looks like the same issue exists with `pairwise` defined in the same file.

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L173

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L173

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L198

https://github.com/JuliaStats/Distances.jl/blob/23b0023cd5129c9e692a557904df1f1ec9088a05/src/generic.jl#L214

dkarrasch:
I started fixing the badges, and soon realized the README could use a little clean-up. Now it mentions the iterator-interface. I wonder whether we should keep the colwise benchmarks. After all, few metrics have a specialized implementation, so we do have a "simple loop implementation", and the speed-up shown results actually from using views in the generic methods. That's something that a user implementing her/his own naive loop might as well have used, so the real benefit from specialized methods is restricted to the [Sq]Mahalanobis and the `CorrDist` metrics. We could shorten the colwise benchmark to those that have a specialized method. Similarly for `pairwise`: the "competitor" doesn't use views. Alternatively, we could include the views in the simple loop competitors, and showcase the results for the specialized BLAS-based methods only?

Closes #204, closes #206.

### DIFF
--- a/README.md
+++ b/README.md
@@ -104,7 +104,7 @@ equally well obtained by
 r = colwise(dist, eachcol(X), eachcol(Y))
 ```
 
-Equally simple, the analogous "rowwise" distance computation can be obtained done via
+Equally simple, the analogous "rowwise" distance computation can be obtained via
 
 ```julia
 r = colwise(dist, eachrow(X), eachrow(Y))
--- a/README.md
+++ b/README.md
@@ -229,7 +229,7 @@ definitions of the distances are listed in the following table.
 |  JSDivergence        |  `js_divergence(p, q)`     | `KL(p, m) / 2 + KL(p, m) / 2 with m = (p + q) / 2` |
 |  SpanNormDist        |  `spannorm_dist(x, y)`     | `max(x - y) - min(x - y)` |
 |  BhattacharyyaDist   |  `bhattacharyya(x, y)`     | `-log(sum(sqrt(x .* y) / sqrt(sum(x) * sum(y)))` |
-|  HellingerDist       |  `hellinger(x, y) `        | `sqrt(1 - sum(sqrt(x .* y) / sqrt(sum(x) * sum(y))))` |
+|  HellingerDist       |  `hellinger(x, y)`        | `sqrt(1 - sum(sqrt(x .* y) / sqrt(sum(x) * sum(y))))` |
 |  Haversine           |  `haversine(x, y, r = 6_371_000)`      | [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) |
 |  SphericalAngle      |  `spherical_angle(x, y)`   | [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) |
 |  Mahalanobis         |  `mahalanobis(x, y, Q)`    | `sqrt((x - y)' * Q * (x - y))` |
--- a/README.md
+++ b/README.md
@@ -256,39 +256,37 @@ replicated using the script in `benchmark/print_table.jl`.
 
 ### Column-wise benchmark
 
-The table below compares the performance (measured in terms of average elapsed
-time of each iteration) of a straightforward loop implementation and an
-implementation specialized to `[Sq]Mahalanobis` provided in *Distances.jl*.
-The task in each iteration is to compute a specific distance between corresponding
-columns in two `200-by-10000` matrices.
+Generically, column-wise distances are computed using a straightforward loop
+implementation. For `[Sq]Mahalanobis`, however, specialized methods are
+provided in *Distances.jl*, and the table below compares the performance
+(measured in terms of average elapsed time of each iteration) of the generic
+to the specialized implementation. The task in each iteration is to compute a
+specific distance between corresponding columns in two `200-by-10000` matrices.
 
-|  distance  |  loop  |  colwise  |  gain  |
-|----------- | -------| ----------| -------|
+|  distance     | loop      |  colwise   |  gain       |
+|---------------|-----------|------------|-------------|
 | SqMahalanobis | 0.089470s |  0.014424s |  **6.2027** |
-| Mahalanobis | 0.090882s |  0.014096s |  **6.4475** |
+| Mahalanobis   | 0.090882s |  0.014096s |  **6.4475** |
 
 ### Pairwise benchmark
 
-The table below compares the performance (measured in terms of average elapsed
-time of each iteration) of a straightforward loop implementation and certain
-specialized implementations provided in *Distances.jl*. The task in each iteration
-is to compute a specific distance in a pairwise manner between columns in a
-`100-by-200` and `100-by-250` matrices, which will result in a `200-by-250`
-distance matrix.
+Generically, pairwise distances are computed using a straightforward loop
+implementation. For distances of which a major part of the computation is a
+quadratic form, however, the performance can be drastically improved by restructuring
+the computation and delegating the core part to `GEMM` in *BLAS*. The table below
+compares the performance (measured in terms of average elapsed time of each
+iteration) of generic to the specialized implementations provided in *Distances.jl*.
+The task in each iteration is to compute a specific distance in a pairwise manner
+between columns in a `100-by-200` and `100-by-250` matrices, which will result in
+a `200-by-250` distance matrix.
 
-|  distance  |  loop  |  pairwise  |  gain  |
-|----------- | -------| ----------| -------|
-| SqEuclidean | 0.001273s |  0.000124s | **10.2290** |
-| Euclidean | 0.001445s |  0.000194s |  **7.4529** |
-| CosineDist | 0.001928s |  0.000149s | **12.9543** |
-| CorrDist | 0.016837s |  0.000187s | **90.1854** |
-| WeightedSqEuclidean | 0.001603s |  0.000143s | **11.2119** |
-| WeightedEuclidean | 0.001811s |  0.000238s |  **7.6032** |
-| SqMahalanobis | 0.308990s |  0.000248s | **1248.1892** |
-| Mahalanobis | 0.313415s |  0.000346s | **906.1836** |
-
-For distances of which a major part of the computation is a quadratic form
-(e.g. *Euclidean*, *CosineDist*, *[Sq]Mahalanobis*), the performance can be
-drastically improved by restructuring the computation and delegating the core
-part to `GEMM` in *BLAS*. The use of this strategy can lead to dramatic
-performance gain over simple loops; see the the table above.
+|  distance             |  loop     |  pairwise  |  gain  |
+|---------------------- | --------- | -----------| -------|
+| SqEuclidean           | 0.001273s |  0.000124s | **10.2290** |
+| Euclidean             | 0.001445s |  0.000194s |  **7.4529** |
+| CosineDist            | 0.001928s |  0.000149s | **12.9543** |
+| CorrDist              | 0.016837s |  0.000187s | **90.1854** |
+| WeightedSqEuclidean   | 0.001603s |  0.000143s | **11.2119** |
+| WeightedEuclidean     | 0.001811s |  0.000238s |  **7.6032** |
+| SqMahalanobis         | 0.308990s |  0.000248s | **1248.1892** |
+| Mahalanobis           | 0.313415s |  0.000346s | **906.1836** |
--- a/README.md
+++ b/README.md
@@ -1,12 +1,13 @@
 # Distances.jl
 
-[![Build Status](https://travis-ci.org/JuliaStats/Distances.jl.svg?branch=master)](https://travis-ci.org/JuliaStats/Distances.jl)
-[![Coverage Status](https://coveralls.io/repos/JuliaStats/Distances.jl/badge.svg?branch=master&service=github)](https://coveralls.io/github/JuliaStats/Distances.jl?branch=master)
+[![Build Status](https://github.com/JuliaStats/Distances.jl/workflows/CI/badge.svg)](https://github.com/JuliaStats/Distances.jl/actions?query=workflow%3ACI)
+[![Coverage Status](http://codecov.io/github/JuliaStats/Distances.jl/coverage.svg?branch=master)](http://codecov.io/github/JuliaStats/Distances.jl?branch=master)
 
 A Julia package for evaluating distances(metrics) between vectors.
 
-This package also provides optimized functions to compute column-wise and pairwise distances, which are often substantially faster than a straightforward loop implementation. (See the benchmark section below for details).
-
+This package also provides optimized functions to compute column-wise and
+pairwise distances, which are often substantially faster than a straightforward
+loop implementation. (See the benchmark section below for details).
 
 ## Supported distances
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -39,99 +40,159 @@ This package also provides optimized functions to compute column-wise and pairwi
...
+The above extends to iterators of iterators/vectors, and the colwise computations
+means the distance computation of corresponding objects (à la `Base.zip`).
+For instance, in Julia v1.1 and above, the colwise behavior shown above can be
+equally well obtained by
+
+```julia
+r = colwise(dist, eachcol(X), eachcol(Y))
+```
+
+Equally simple, the analogous "rowwise" distance computation can be obtained done via
```
devmotion, line 107:
```suggestion
Equally simple, the analogous "rowwise" distance computation can be obtained via
```

devmotion:
In my opinon, it is a bit strange and misleading to benchmark against an implememtation that does not use `view`s.

dkarrasch:
> In my opinon, it is a bit strange and misleading to benchmark against an implememtation that does not use `view`s.

I agree. I'll go ahead and reduce the benchmark sections to only those methods that have specialized col- or pairwise implementations.

dkarrasch:
Just to reassure everybody, here are the full benchmark tables computed on this branch:
```
|  distance  |  loop  |  colwise  |  gain  |
|----------- | -------| ----------| -------|
| SqEuclidean | 0.001472s |  0.001408s |  1.0455 |
| Euclidean | 0.001482s |  0.001417s |  1.0460 |
| PeriodicEuclidean | 0.007600s |  0.007529s |  1.0094 |
| Cityblock | 0.001505s |  0.001444s |  1.0424 |
| TotalVariation | 0.001507s |  0.001429s |  1.0546 |
| Chebyshev | 0.005383s |  0.005310s |  1.0138 |
| Minkowski | 0.044728s |  0.043700s |  1.0235 |
| Hamming | 0.001500s |  0.001424s |  1.0538 |
| CosineDist | 0.001518s |  0.001445s |  1.0505 |
| CorrDist | 0.005794s |  0.006070s |  0.9546 |
| ChiSqDist | 0.001701s |  0.001657s |  1.0267 |
| KLDivergence | 0.027774s |  0.028905s |  0.9609 |
| RenyiDivergence | 0.011712s |  0.011653s |  1.0051 |
| RenyiDivergence | 0.008443s |  0.007358s |  1.1475 |
| JSDivergence | 0.044401s |  0.044594s |  0.9957 |
| BhattacharyyaDist | 0.003822s |  0.003820s |  1.0004 |
| HellingerDist | 0.003819s |  0.003816s |  1.0007 |
| WeightedSqEuclidean | 0.001500s |  0.001449s |  1.0352 |
| WeightedEuclidean | 0.001511s |  0.001451s |  1.0410 |
| WeightedCityblock | 0.001508s |  0.001456s |  1.0354 |
| WeightedMinkowski | 0.044505s |  0.045645s |  0.9750 |
| WeightedHamming | 0.001481s |  0.001437s |  1.0308 |
| SqMahalanobis | 0.088364s |  0.013481s |  6.5548 |
| Mahalanobis | 0.087668s |  0.013698s |  6.3998 |
| BrayCurtis | 0.001590s |  0.001598s |  0.9950 |

|  distance  |  loop  |  pairwise  |  gain  |
|----------- | -------| ----------| -------|
| SqEuclidean | 0.001262s |  0.000127s |  9.9500 |
| Euclidean | 0.001438s |  0.000200s |  7.1934 |
| PeriodicEuclidean | 0.019845s |  0.019593s |  1.0129 |
| Cityblock | 0.001286s |  0.000878s |  1.4657 |
| TotalVariation | 0.001354s |  0.000929s |  1.4576 |
| Chebyshev | 0.013168s |  0.012528s |  1.0511 |
| Minkowski | 0.117622s |  0.112400s |  1.0465 |
| Hamming | 0.001145s |  0.000716s |  1.5987 |
| CosineDist | 0.001909s |  0.000142s | 13.4772 |
| CorrDist | 0.015215s |  0.000179s | 84.8529 |
| ChiSqDist | 0.003212s |  0.003210s |  1.0006 |
| KLDivergence | 0.069010s |  0.067663s |  1.0199 |
| RenyiDivergence | 0.028517s |  0.027525s |  1.0360 |
| RenyiDivergence | 0.020241s |  0.019774s |  1.0236 |
| JSDivergence | 0.117897s |  0.117746s |  1.0013 |
| BhattacharyyaDist | 0.009727s |  0.009825s |  0.9900 |
| HellingerDist | 0.009293s |  0.009255s |  1.0042 |
| WeightedSqEuclidean | 0.001593s |  0.000138s | 11.5449 |
| WeightedEuclidean | 0.001845s |  0.000232s |  7.9398 |
| WeightedCityblock | 0.001709s |  0.001337s |  1.2782 |
| WeightedMinkowski | 0.115608s |  0.119167s |  0.9701 |
| WeightedHamming | 0.001575s |  0.001229s |  1.2816 |
| SqMahalanobis | 0.301555s |  0.000254s | 1187.0757 |
| Mahalanobis | 0.302782s |  0.000343s | 883.4002 |
| BrayCurtis | 0.001858s |  0.001515s |  1.2266 |
```
Significant deviations in "gain" from 1 start at 5, and are exclusively related to specialized methods. Everything else seems to be pretty close to 1.

README.md
```diff
@@ -40,99 +41,159 @@ This package also provides optimized functions to compute column-wise and pairwi
...
+equally well obtained by
+
+```julia
+r = colwise(dist, eachcol(X), eachcol(Y))
+```
+
+Equally simple, the analogous "rowwise" distance computation can be obtained via
+
+```julia
+r = colwise(dist, eachrow(X), eachrow(Y))
```
nalimilan, line 111: This sounds terribly confusing. We should use `map` for this instead. I'd avoid mentioning this until we have a more logical story. :-)

README.md
```diff
@@ -167,8 +229,8 @@ Each distance corresponds to a distance type. The type name and the correspondin
 |  JSDivergence        |  `js_divergence(p, q)`     | `KL(p, m) / 2 + KL(p, m) / 2 with m = (p + q) / 2` |
 |  SpanNormDist        |  `spannorm_dist(x, y)`     | `max(x - y) - min(x - y)` |
 |  BhattacharyyaDist   |  `bhattacharyya(x, y)`     | `-log(sum(sqrt(x .* y) / sqrt(sum(x) * sum(y)))` |
-|  HellingerDist       |  `hellinger(x, y) `        | `sqrt(1 - sum(sqrt(x .* y) / sqrt(sum(x) * sum(y))))` |
-|  Haversine           |  `haversine(x, y, r = 6371000)`      | [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) |
+|  HellingerDist       |  `hellinger(x, y)`        | `sqrt(1 - sum(sqrt(x .* y) / sqrt(sum(x) * sum(y))))` |
+|  Haversine           |  `haversine(x, y, r = 6_371_000)`      | [Haversine formula](https://en.wikipedia.org/wiki/Haversine_formula) |
```
nalimilan, line 233: While you're at it, maybe realign the columns with spaces?

README.md
```diff
@@ -40,99 +41,159 @@ This package also provides optimized functions to compute column-wise and pairwi
...
-`pairwise(dist, X)` is typically more efficient than `pairwise(dist, X, X)`, as the former will take advantage of the symmetry when `dist` is a semi-metric (including metric).
+This statement will result in an `m-by-m` matrix, where `R[i,j]` is the distance
+between `X[:,i]` and `X[:,j]`. `pairwise(dist, X)` is typically more efficient
+than `pairwise(dist, X, X)`, as the former will take advantage of the symmetry
+when `dist` is a semi-metric (including metric).
+
+For performance reasons, it is recommended to use matrices with observations in
+columns (as shown above). Indeed, the `Array` type in Julia is column-major,
+making it more efficient to access memory column by column. However, matrices
+with observations stored in rows are also supported via the argument `dims=1`.
```
nalimilan, line 149: BTW, I've never been able to reproduce this alleged performance difference. Do you have any idea how we could reproduce it?

README.md
```diff
@@ -40,99 +41,159 @@ This package also provides optimized functions to compute column-wise and pairwi
...
+
+```julia
+r = colwise(dist, eachrow(X), eachrow(Y))
+```
 
-Note that either of `X` and `Y` can be just a single vector -- then the `colwise` function will compute the distance between this vector and each column of the other parameter.
+The iterator-based approach is in fact the future-proof way of computing distances
+between corresponding data points of iterable collections `X` and `Y`. Exceptions
+are given by distance types with `colwise` methods specialized for
+`AbstractVecOrMat` arguments, i.e., `CorrDist` and `[Sq]Mahalanobis`.
```
nalimilan, line 117: The wording makes it sound like for these distances the iterator-based approach wouldn't be future-proof. But the problem is rather than they are currently slower. Same below for `pairwise`.

src/metrics.jl
```diff
@@ -720,33 +721,34 @@ function _pairwise!(r::AbstractMatrix, dist::Euclidean,
...
     if threshT <= 0
+        # If there's no chance of triggering the threshold, we can use @simd
         for j = 1:nb
             sb = sb2[j]
             @simd for i = 1:na
                  @inbounds r[i, j] = sqrt(max(sa2[i] + sb - 2 * r[i, j], 0))
             end
         end
     else
-        @inbounds for j = 1:nb
```
nalimilan, line 734: Why change this?

src/metrics.jl
```diff
@@ -786,12 +788,6 @@ function _pairwise!(r::AbstractMatrix, dist::Euclidean, a::AbstractMatrix)
 end
 
 # Weighted Euclidean
-function colwise!(r::AbstractArray, dist::WeightedEuclidean, a::AbstractMatrix, b::AbstractMatrix)
-    sqrt!(colwise!(r, WeightedSqEuclidean(dist.weights), a, b))
-end
-function colwise!(r::AbstractArray, dist::WeightedEuclidean, a::AbstractVector, b::AbstractMatrix)
-    sqrt!(colwise!(r, WeightedSqEuclidean(dist.weights), a, b))
-end
```
nalimilan, line 794: I understand these are redundant with the fallback methods? Does the fallback also avoids making a copy like `sqrt!`?

README.md
dkarrasch: No idea. We could leave out the recommendation part and simply mention the row case, without judgement.

src/metrics.jl
dkarrasch: The fallback simply computes the `sqrt` in the course of `evaluate` (in `eval_end`). The fallback might be even slightly faster, because it doesn't loop over the result array twice.

src/metrics.jl
dkarrasch: We had slightly different implementations for `SqEuclidean` and `Euclidean`, and the former was faster. Now that I think about it again, I'll reduce the code duplication using `eval_end`.

README.md
```diff
@@ -1,12 +1,13 @@
 # Distances.jl
 
-[![Build Status](https://travis-ci.org/JuliaStats/Distances.jl.svg?branch=master)](https://travis-ci.org/JuliaStats/Distances.jl)
-[![Coverage Status](https://coveralls.io/repos/JuliaStats/Distances.jl/badge.svg?branch=master&service=github)](https://coveralls.io/github/JuliaStats/Distances.jl?branch=master)
+[![Build Status](https://github.com/JuliaStats/Distances.jl/workflows/CI/badge.svg)](https://github.com/JuliaStats/Distances.jl/actions?query=workflow%3ACI)
```
devmotion, line 3: This would ensure that only results of the master branch are displayed and linked to:
```suggestion
[![Build Status](https://github.com/JuliaStats/Distances.jl/workflows/CI/badge.svg?branch=master)](https://github.com/JuliaStats/Distances.jl/actions?query=workflow%3ACI+branch%3Amaster)
```

src/metrics.jl
```diff
@@ -660,27 +661,27 @@ function _pairwise!(r::AbstractMatrix, dist::SqEuclidean, a::AbstractMatrix)
...
+            selfterms = sa2[i] + sa2j
+            v = max(selfterms - 2 * r[i, j], 0)
+            if v < threshT * selfterms
+                v = zero(v)
+                for k = 1:m
+                    v += (a[k, i] - a[k, j])^2
                 end
-                r[i, j] = v
+            end
+            r[i, j] = eval_end(dist, v)
```
nalimilan, line 676: Unintended change?

README.md
```diff
@@ -212,79 +244,51 @@ julia> pairwise(Euclidean(1e-12), x, x)
...
 
-#### Column-wise benchmark
+### Column-wise benchmark
 
-The table below compares the performance (measured in terms of average elapsed time of each iteration) of a straightforward loop implementation and an optimized implementation provided in *Distances.jl*. The task in each iteration is to compute a specific distance between corresponding columns in two `200-by-10000` matrices.
+The table below compares the performance (measured in terms of average elapsed
+time of each iteration) of a straightforward loop implementation and an
+implementation specialized to `[Sq]Mahalanobis` provided in *Distances.jl*.
+The task in each iteration is to compute a specific distance between corresponding
+columns in two `200-by-10000` matrices.
```
nalimilan, line 263: How about explaining that for other distances the implementation is equivalent to a loop?

README.md
```diff
@@ -212,79 +244,51 @@ julia> pairwise(Euclidean(1e-12), x, x)
...
+| WeightedSqEuclidean | 0.001603s |  0.000143s | **11.2119** |
+| WeightedEuclidean | 0.001811s |  0.000238s |  **7.6032** |
+| SqMahalanobis | 0.308990s |  0.000248s | **1248.1892** |
+| Mahalanobis | 0.313415s |  0.000346s | **906.1836** |
+
+For distances of which a major part of the computation is a quadratic form
+(e.g. *Euclidean*, *CosineDist*, *[Sq]Mahalanobis*), the performance can be
+drastically improved by restructuring the computation and delegating the core
+part to `GEMM` in *BLAS*. The use of this strategy can lead to dramatic
+performance gain over simple loops; see the the table above.
```
nalimilan, line 294: It would be more logical to put this before the table IMO. And again mention that for other distances the naive approach is used.

README.md
nalimilan: Yeah I'm not sure what's best. One day we should make more serious benchmarks to resolve this question. :-)

src/metrics.jl
dkarrasch: No, this is fully intentional. I have merged the two codes for pairwise Euclidean and SqEuclidean, which differed only in the final sqrt (and in a different placement of `@inbounds`. The diff log may look a bit weird though.

dkarrasch:
I have rearranged the statements in the benchmark section, and used the opportunity to bump the patch version right away.

src/metrics.jl
nalimilan: Sorry, what I mean is that the indentation is wrong.

dkarrasch: 🤦

nalimilan (approved):

PR #203, opened 2021-01-11
Allow unitful data points

### LINKED ISSUE #201: Euclidean distance does not work with Unitful vectors
Exploring the [Unitful](https://github.com/PainterQubits/Unitful.jl) package I noticed that it doesn't play well with Distances – attempting to compute the Euclidean distance between two vectors whose elements are in meters resulted in an error.

The Euclidean distance of two vectors in meters is a well-defined operation and computing it naively using functions included in Base works fine:

```julia
julia> using Unitful.DefaultSymbols, Distances

julia> x = [1m, 2m, 3m]
3-element Array{Unitful.Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}:
 1 m
 2 m
 3 m

julia> y = [4m, 5m, 6m]
3-element Array{Unitful.Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}:
 4 m
 5 m
 6 m

julia> sqrt(sum(abs2.(x - y)))
5.196152422706632 m
```

However, using Distances to do the computation fails:

```julia
julia> euclidean(x, y)
ERROR: DimensionError: 0.0 m and 9 m² are not dimensionally compatible.
Stacktrace:
 [1] +(::Unitful.Quantity{Float64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}, ::Unitful.Quantity{Int64,𝐋²,Unitful.FreeUnits{(m²,),𝐋²,nothing}}) at /Users/yurivish/.julia/packages/Unitful/1t88N/src/quantities.jl:137
 [2] eval_reduce(::Euclidean, ::Unitful.Quantity{Float64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}}, ::Unitful.Quantity{Int64,𝐋²,Unitful.FreeUnits{(m²,),𝐋²,nothing}}) at /Users/yurivish/.julia/packages/Distances/1CdE2/src/metrics.jl:295
 [3] macro expansion at /Users/yurivish/.julia/packages/Distances/1CdE2/src/metrics.jl:243 [inlined]
 [4] macro expansion at ./simdloop.jl:77 [inlined]
 [5] _evaluate at /Users/yurivish/.julia/packages/Distances/1CdE2/src/metrics.jl:240 [inlined]
 [6] Euclidean at /Users/yurivish/.julia/packages/Distances/1CdE2/src/metrics.jl:299 [inlined]
 [7] euclidean(::Array{Unitful.Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}, ::Array{Unitful.Quantity{Int64,𝐋,Unitful.FreeUnits{(m,),𝐋,nothing}},1}) at /Users/yurivish/.julia/packages/Distances/1CdE2/src/metrics.jl:308
 [8] top-level scope at REPL[5]:1
```

I debugged this somewhat and believe that the bug has to do with the fact that intermediate values used in the computation of the Euclidean distance do not share the same units as the result. The `abs2` operation in my naive implementation results in an intermediate vector whose values are in meters squared:

```julia
julia> abs2.(x - y)
3-element Array{Unitful.Quantity{Int64,𝐋²,Unitful.FreeUnits{(m²,),𝐋²,nothing}},1}:
 9 m²
 9 m²
 9 m²
```

But the definition of `_evaluate` in [metrics.jl](https://github.com/JuliaStats/Distances.jl/blob/f6ee3532b2c312da80692ec3d84ca796e45d432d/src/metrics.jl#L242) accumulates the result into `s` through a series of reductions with the intial value of `s` set to be a zero of the result type using [`eval_start`](https://github.com/JuliaStats/Distances.jl/blob/f6ee3532b2c312da80692ec3d84ca796e45d432d/src/metrics.jl#L317).

The value `s` begins life as `0 m` and the reduction step adds this to the intermediate value `9 m²`, which throws the `DimensionError`:

https://github.com/JuliaStats/Distances.jl/blob/f6ee3532b2c312da80692ec3d84ca796e45d432d/src/metrics.jl#L255

dkarrasch:
Closes #201. Indeed, there is no reason why in general the temporary variable should have the same type/unit as the final result. Funnily, this added flexibility leads to a significant gain in performance in special cases where the types differ:
```julia
julia> using Distances, BenchmarkTools

julia> x = rand(1:5, 100);

julia> y = rand(1:5, 100);

julia> @btime euclidean($x, $y);
  95.578 ns (0 allocations: 0 bytes) # master
  33.965 ns (0 allocations: 0 bytes) # this PR
```
We need to mimick the `result_type` scheme once more, but the added code is not much. Pretty much all the metrics are covered, except those that involve a `log`, the `Mahalanobis` family, `Bregman`. ~~I need to check if `haversine` works with `°`.~~ EDIT: it does now.

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -11,34 +11,27 @@ The computed distance has the unit of the radius.
 struct Haversine{T} <: Metric
     radius::T
 end
-Haversine() = Haversine(Float32(6371000))
+Haversine() = Haversine(Float32(6_371_000))
 
 function (dist::Haversine)(x, y)
     length(x) == length(y) == 2 || haversine_error(dist)
 
-    @inbounds x1, x2 = x
-    @inbounds y1, y2 = y
-    # longitudes
-    Δλ = _deg2rad(y1 - x1)
+    @inbounds λ₁, φ₁ = x
+    @inbounds λ₂, φ₂ = y
 
-    # latitudes
-    φ₁ = _deg2rad(x2)
-    φ₂ = _deg2rad(y2)
-    Δφ = φ₂ - φ₁
+    Δλ = λ₂ - λ₁  # longitudes
+    Δφ = φ₂ - φ₁  # latitudes
 
     # haversine formula
-    a = sin(Δφ/2)^2 + cos(φ₁)*cos(φ₂)*sin(Δλ/2)^2
+    a = sind(Δφ/2)^2 + cosd(φ₁)*cosd(φ₂)*sind(Δλ/2)^2
 
     # distance on the sphere
     2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
 end
 
-haversine(x, y, radius=Float32(6371000)) = Haversine(radius)(x, y)
+haversine(x, y, radius=Float32(6_371_000)) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
-
-_deg2rad(x::Real) = deg2rad(x)
-_deg2rad(x::Number) = x # targets unitful numbers
 
 """
     SphericalAngle()
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -2,7 +2,8 @@
     Haversine([radius])
 
 The haversine distance between two locations on a sphere of given `radius`, whose
-default value is 6371000, i.e., the Earth's mean radius in units of meter.
+default value is 6371000, i.e., the Earth's (volumetric) mean radius in units of meter; cf.
+[NASA's Earth Fact Sheet](https://nssdc.gsfc.nasa.gov/planetary/factsheet/earthfact.html).
 
 Locations are described with longitude and latitude in degrees.
 The computed distance has the unit of the radius.
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -8,7 +8,7 @@ default value is 6,371,000, i.e., the Earth's (volumetric) mean radius in meters
 Locations are described with longitude and latitude in degrees.
 The computed distance has the unit of the radius.
 """
-struct Haversine{T} <: Metric
+struct Haversine{T<:Number} <: Metric
     radius::T
 end
 Haversine() = Haversine(Float32(6_371_000))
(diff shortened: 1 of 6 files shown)


dkarrasch:
I've now modified the `Haversine` computation slightly to also handle arguments with degree units properly. While I was at it, I added the default radius and references & documentation.

dkarrasch:
This is ready for review/merge.

src/haversine.jl
```diff
@@ -1,14 +1,14 @@
 """
     Haversine([radius])
 
-The haversine distance between two locations on a sphere of given `radius`.
+The haversine distance between two locations on a sphere of given `radius`, whose
+default value is 6371000, i.e., the Earth's (volumetric) mean radius in units of meter; cf.
```
nalimilan, line 5:
```suggestion
default value is 6,371,000, i.e., the Earth's (volumetric) mean radius in meters; cf.
```

src/haversine.jl
```diff
@@ -1,14 +1,14 @@
 """
     Haversine([radius])
```
nalimilan, line 2:
```suggestion
    Haversine(radius=6_371_000)
```

src/haversine.jl
```diff
@@ -33,11 +33,12 @@ function (dist::Haversine)(x, y)
...
 end
 
-haversine(x, y, radius::Real = Float32(6371000)) = Haversine(radius)(x, y)
+haversine(x, y, radius=Float32(6371000)) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 
-
+_deg2rad(x::Real) = deg2rad(x)
+_deg2rad(x) = x # targets unitful numbers
```
nalimilan, line 41: Maybe at least continue requiring `::Number`?

Also, is this the only place where we required `::Real`?

dkarrasch: The only other hits for `::Real` I found are `p` in `*Minkowski` and `alpha` in `RenyiDivergence`, which both seem reasonable to me.

src/haversine.jl
```diff
@@ -33,11 +33,12 @@ function (dist::Haversine)(x, y)
...
 end
 
-haversine(x, y, radius::Real = Float32(6371000)) = Haversine(radius)(x, y)
+haversine(x, y, radius=Float32(6371000)) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 
-
+_deg2rad(x::Real) = deg2rad(x)
+_deg2rad(x::Number) = x # targets unitful numbers
```
nalimilan, line 41: Actually, couldn't Unitful define `deg2rad` when the unit is degree? Otherwise, what happens if somebody passes a radius e.g. in kg?

dkarrasch: I think I found a much better way to do it, thanks for persisting and pushing. ;-)

nalimilan: Great! Maybe we should keep the `::Number` restriction?

nalimilan (approved):

PR #219, opened 2021-05-31
correct _pairwise! function for complex input

wasyro:
`pairwise` function with `Euclidean`, `SqEuclidean`, `WeightedEuclidean`, or `WeightedSqEuclidean` doesn't work when a complex-valued matrix is passed.
```
julia> Y = rand(10, 100) .+ im .* rand(10, 100)
julia> pairwise(SqEuclidean(), Y)

ERROR: InexactError: Float64(4.879925878421758 + 0.6223517426063204im)
Stacktrace:
  [1] Real
    @ ./complex.jl:37 [inlined]
  [2] convert
    @ ./number.jl:7 [inlined]
  [3] setindex!
    @ ./array.jl:841 [inlined]
  [4] setindex!
    @ ./multidimensional.jl:639 [inlined]
  [5] _modify!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/generic.jl:91 [inlined]
  [6] _generic_matmatmul!(C::Matrix{Float64}, tA::Char, tB::Char, A::Matrix{ComplexF64}, B::Matrix{ComplexF64}, _add::LinearAlgebra.MulAddMul{true, true, Bool, Bool})
    @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:955
  [7] generic_matmatmul!(C::Matrix{Float64}, tA::Char, tB::Char, A::Matrix{ComplexF64}, B::Matrix{ComplexF64}, _add::LinearAlgebra.MulAddMul{true, true, Bool, Bool})
    @ LinearAlgebra /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:802
  [8] mul!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:458 [inlined]
  [9] mul!
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/LinearAlgebra/src/matmul.jl:275 [inlined]
 [10] _pairwise!(r::Matrix{Float64}, dist::SqEuclidean, a::Matrix{ComplexF64})
    @ Distances ~/.julia/dev/Distances/src/metrics.jl:654
[...]

PR #220, opened 2021-05-31
Allow complex arrays in *Euclidean pairwise

dkarrasch:
Alternative to #219, without code duplication.

Closes #219.

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -613,7 +613,8 @@ nrmsd(a, b) = NormRMSDeviation()(a, b)
 function _pairwise!(r::AbstractMatrix, dist::Union{SqEuclidean,Euclidean},
                     a::AbstractMatrix, b::AbstractMatrix)
     m, na, nb = get_pairwise_dims(r, a, b)
-    mul!(r, a', b)
+    inplace = promote_type(eltype(r), typeof(oneunit(eltype(a))'oneunit(eltype(b)))) === eltype(r)
+    R = inplace ? mul!(r, a', b) : a'b
     sa2 = sum(abs2, a, dims=1)
     sb2 = sum(abs2, b, dims=1)
     threshT = convert(eltype(r), dist.thresh)
@@ -622,7 +623,7 @@ function _pairwise!(r::AbstractMatrix, dist::Union{SqEuclidean,Euclidean},
         for j = 1:nb
             sb = sb2[j]
             @simd for i = 1:na
-                @inbounds r[i, j] = eval_end(dist, (max(sa2[i] + sb - 2 * r[i, j], 0)))
+                @inbounds r[i, j] = eval_end(dist, (max(sa2[i] + sb - 2real(R[i, j]), 0)))
             end
         end
     else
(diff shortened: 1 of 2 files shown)


dkarrasch:
Anyone who would like to take a look? I've also included some `require_one_based_indexing` checks which used to be implicit but always necessary (otherwise matrix multiplications wouldn't work). Is that fine or should I do that separately to better track individual changes?

src/metrics.jl
```diff
@@ -612,45 +612,49 @@ nrmsd(a, b) = NormRMSDeviation()(a, b)
 # SqEuclidean/Euclidean
 function _pairwise!(r::AbstractMatrix, dist::Union{SqEuclidean,Euclidean},
                     a::AbstractMatrix, b::AbstractMatrix)
+    require_one_based_indexing(r, a, b)
     m, na, nb = get_pairwise_dims(r, a, b)
-    mul!(r, a', b)
+    inplace = promote_type(eltype(r), typeof(oneunit(eltype(a))'oneunit(eltype(b)))) === eltype(r)
```
johnnychen94, line 617: Leave a comment for this might be helpful for future readers.

johnnychen94 (approved):
*__looks__* good to me (I don't use complex numbers).

juliohm:
I wonder if this generalization could also work with color types? For example pairwise between Array{Gray}. I missed this functionality yesterday trying to cluster a collection of pixels with a method that calls `pairwise(Euclidean(), colors, colors)`.

johnnychen94:
No, because `abs2` is undefined for most colorants. More over, in ColorVectorSpace v0.10 release `abs2(::RGB)` is removed.

> I missed this functionality yesterday trying to cluster a collection of pixels with a method that calls pairwise(Euclidean(), colors, colors).

Isn't `ImageDistances` designed for this purpose?

juliohm:
In the context of general machine learning some features are colors and
some others are not. So the clustering algorithm we have in GeoStats.jl
needs to call a generic pairwise distance not just image channels as
features. It would be nice to have it work in general. You mean that not
even Gray has a meaningful distance definition?

johnnychen94:
> You mean that not even `Gray` has a meaningful distance definition?

See discussion in https://github.com/JuliaGraphics/ColorVectorSpace.jl/issues/157

And `pairwise(Euclidean(), colors, colors)` is already supported by ImageDistances:

```julia
julia> using ImageDistances

julia> x = rand(Gray{Float32}, 16, 2);

julia> pairwise(Euclidean(), x, x)
2×2 Matrix{Float32}:
 0.0      1.73782
 1.73782  0.0

julia> pairwise(Euclidean(), x, Float32.(x)) # mixed type
2×2 Matrix{Float32}:
 0.0      1.73782
 1.73782  0.0
```

This is off-topic so here's the last clarification in this PR; if you have further specific concerns please open issues in either ColorVectorSpace or ImageDistances.

juliohm:
I find it suboptimal to depend on a separate package to handle this use case. ImageDistances.jl was created with a different purpose in mind. Namely, to define truly spatial distances that look at more than one pixel at a time like the Hausdorff distance. It was not created to be a general approach to compute distance between colors.

It would be interesting to revive this discussion in some other issue later. The original design in Images.jl shouldn´t be forgot where an image is simply an Array of color objects. If we keep developing image-specific packages like ImageDistances.jl with a different picture in mind, we will create an isolated ecosystem that only works in specific use cases. The discussion around abs2 is controversial with many different opinions and we should reconsider a generic treatment of colors if that is possible.

Again I started ImageDistances.jl with a very specific set of distances in mind, the image-specific distances that cannot be expressed in terms of pixel-wise operations.

dkarrasch:
Since this is already adding functionality, shall we have a separate discussion about the image-related issue? The main question seems to be what to do about the `abs2` colwise reduction, right? One possible solution could be to replace that by some `_abs2`, which defaults to `abs2` but could be overloaded by other packages for custom types.

Since there seem to be no objections, I'll go and merge this PR, but wait for other recent PRs before a release.

PR #221, opened 2021-05-31
Require symmetry and check isposdef in `*Mahalanobis`

### LINKED ISSUE #174: Check that the matrix of (Sq)Mahalanobis is positive-definite?
Currently, upon creation of `(Sq)Mahalanobis(Q)` it is not checked if `Q` is actually positive (semi-)definite, and hence in particular symmetric, as shown in the following example:
```julia
julia> x = rand(2);

julia> sqmahalanobis(x, -x, Matrix(-I, 2, 2))
-1.0767684441054513

julia> mahalanobis(x, -x, Matrix(-I, (2, 2)))
ERROR: DomainError with -1.0767684441054513:
sqrt will only return a complex result if called with a complex argument. Try sqrt(Complex(x)).
Stacktrace:
 [1] throw_complex_domainerror(::Symbol, ::Float64) at ./math.jl:33
 [2] sqrt at ./math.jl:573 [inlined]
 [3] Mahalanobis at /home/david/.julia/dev/Distances/src/mahalanobis.jl:88 [inlined]
 [4] mahalanobis(::Array{Float64,1}, ::Array{Float64,1}, ::Array{Int64,2}) at /home/david/.julia/dev/Distances/src/mahalanobis.jl:91
 [5] top-level scope at REPL[33]:1
```

However, the implementation of `pairwise!(::SqMahalanobis, X, Y)` is based on the identity `(x-y)'*Q*(x-y) = x'*Q*x + y'*Q*y - x'*Q*y - y'*Q*x = x'*Q*x + y'*Q*y - 2 * x'*Q*y` which holds only if `Q` is symmetric.
This leads to the surprising behaviour
```julia
julia> X = rand(2, 3);

julia> Y = rand(2, 4);

julia> Q = rand(2, 2);

julia> pairwise(SqMahalanobis(Q), X, Y; dims=2)
3×4 Array{Float64,2}:
 -0.0807376   0.0288264  -0.0246195  0.00500487
 -0.00275356  0.0291969   0.088852   0.0368224
 -0.0131345   0.0303736   0.0619867  0.0329054

julia> pairwise(SqMahalanobis(Array(Q')), X, Y; dims=2)
3×4 Array{Float64,2}:
 0.0284595   -0.0490699  0.0981837   0.000274926
 0.00392172  -0.154261   0.0615561  -0.0740337
 0.0106349   -0.144796   0.0579041  -0.0660179

julia> map(Iterators.product(eachcol(X), eachcol(Y))) do (x, y)
           z = x - y
           dot(z, Q, z)
       end
3×4 Array{Float64,2}:
 -0.026139     -0.0101218  0.0367821   0.0026399
  0.000584081  -0.0625319  0.0752041  -0.0186056
 -0.0012498    -0.0572112  0.0599454  -0.0165563

julia> map(Iterators.product(eachcol(X), eachcol(Y))) do (x, y)
           z = x - y
           dot(z, Q', z)
       end
3×4 Array{Float64,2}:
 -0.026139     -0.0101218  0.0367821   0.0026399
  0.000584081  -0.0625319  0.0752041  -0.0186056
 -0.0012498    -0.0572112  0.0599454  -0.0165563
```
although
```julia
julia> sqmahalanobis(X[:, 1], Y[:, 1], Q)
-0.026139038969147838

julia> sqmahalanobis(X[:, 1], Y[:, 1], Array(Q'))
-0.026139038969147838
```

Hence, due to the missing check and the inconsistencies in `pairwise!`, I am a bit worried that users might not be aware of these assumptions and their impact on `pairwise!` and might silently obtain incorrect values (see also https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/154#discussion_r474772998).

Maybe it would be best to check if the matrix of (Sq)Mahalanobis is positive-definite in the constructor? Or at least mention this assumption prominently in the README? Alternatively, one could neglect the mathematical motivation completely and make the results consistent by not assuming symmetry in the `pairwise!` computations, i.e., one could just calculate `x'*Q*x + y'*Q*y - x'*Q*y - y'*Q*x`.

dkarrasch:
Closes #174.


src/mahalanobis.jl
```diff
@@ -2,11 +2,31 @@
...
+Mahalanobis(Q::AbstractMatrix; isposdef::Bool = isposdef(Q)) =
+    Mahalanobis{typeof(Q)}(Q, isposdef)
 
 struct SqMahalanobis{M<:AbstractMatrix} <: SemiMetric
     qmat::M
+    _isposdef::Bool
+    function SqMahalanobis{M}(Q::M, isposdef::Bool) where {M <: AbstractMatrix}
+        if !isposdef
+            ishermitian(Q) || throw(ArgumentError("bilinear form is not symmetric/Hermitian"))
+            @warn "bilinear form is not positive definite"
```
devmotion, line 23:
```suggestion
            @warn "bilinear form is not positive definite"
```

src/mahalanobis.jl
```diff
@@ -2,11 +2,31 @@
 
 struct Mahalanobis{M<:AbstractMatrix} <: Metric
     qmat::M
+    _isposdef::Bool
+    function Mahalanobis{M}(Q::M, isposdef::Bool) where {M <: AbstractMatrix}
+        if !isposdef
+            ishermitian(Q) || throw(ArgumentError("bilinear form is not symmetric/Hermitian"))
+            @warn "bilinear form is not positive definite"
```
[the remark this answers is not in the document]
devmotion: However, to be honest I think it is strange to print a warning here. I assume the motivation is to warn users that provide non-positive definite matrices without noticing but with this approach also users that specify `isposdef=false` explicitly will get a warning.

devmotion (commented):
I am not convinced yet that this is the right approach to fix https://github.com/JuliaStats/Distances.jl/issues/174. I see some problems here:
- only positive semi-definiteness is required by the current implementation but this approach checks positive definiteness
- a warning is displayed even if users specify explicitly `isposdef=false` or the matrix is positive-semidefinite
- one can't opt out of the symmetry check which could be desired as well (workaround: use `Symmetry(matrix)` wrapper)
- matrices that are not positive semi-definite don't yield a proper `Metric` (mathematically, positive semi-definite matrices don't yield a metric but they still satisfy all properties of `Metric`)

So I wonder if instead one should maybe remove the `_isposdef` field and only support a `ispossemidef` keyword argument with default `isposdef(Q)` and always error if it is `false`. The only annoying thing would be that `isposdef(Q)` would return `false` also for positive semi-definite matrices so users have to set `ispossemidef=true` explicitly. Maybe that's  a reasonable trade-off to avoid surprising results such as in the issue.

devmotion:
> The only annoying thing would be that `isposdef(Q)` would return `false` also for positive semi-definite matrices so users have to set `ispossemidef=true` explicitly.

Maybe one could use `eigmin(Q) >= 0` instead of `isposdef(Q)`.

src/mahalanobis.jl
```diff
@@ -2,10 +2,24 @@
 
 struct Mahalanobis{M<:AbstractMatrix} <: Metric
     qmat::M
+    issemiposdef::Bool
```
devmotion, line 5: I still think there is no need to store this in a field and to adapt the algorithms since if it is not positive semi-definite then `Mahalanobis` and `SqMahalanobis` are not even `PreMetric`s.

In my opinion, one should just add a constructor that allows to skip the check, e.g., by defining
```julia
function Mahalanobis(Q::AbstractMatrix; issemiposdef=eigmin(Q) >= 0)
    ishermitian(Q) || throw(ArgumentError("bilinear form is not symmetric/Hermitian"))
    issemiposdef || throw(ArgumentError("bilinear form is not positive semidefinite"))
    return Mahalanobis{typeof(Q)}(Q)
end
```

dkarrasch: I thought there was a use case for indefinite matrices? https://github.com/JuliaStats/Distances.jl/issues/174#issuecomment-710688804 If we enforce both properties, then surely we don't need the field and the branches in the code.

devmotion: I think both comments mainly complained that the assumptions are not clearly enforced and hence it is easy to end up with nonsense results if one does not provide a proper positive semidefinite matrix. At least I don't have a use case for it and also the other comment mentioned that it is mainly an error on the user side.

dkarrasch: Then how about making it an optional argument `skipchecks`, make it default to `false` and skip both conditions if `skipchecks == true`? Then we avoid having to memorize how the argument is called.

dkarrasch: Ahh, that doesn't work because it overloads the standard constructor...

devmotion: Is there anything you dislike about the suggestion above? It seems reasonably explicit and direct and it would still allow one to skip all checks by just calling the inner constructor if desired.

dkarrasch: How do you call the inner constructor? `Mahalanobis(Q)` will append the default value for `issemiposdef`.

devmotion: `Mahalanobis{typeof(Q)}(Q)`.

dkarrasch: It's not that I dislike it, it's just that you have to remember the name of the kwarg, and I find it ugly to ask the user to type `Mahalanobis{typeof(Q)}(Q)` to avoid checks. What do you think of the approach in fa38714?

devmotion: I think it would be clearer with keyword arguments, it's not completely obvious what `true` in `Mahalanobis(M, true)` means.

src/mahalanobis.jl
```diff
@@ -1,11 +1,67 @@
...
+ERROR: ArgumentError: bilinear form is not symmetric/Hermitian
+
+julia> dist = Mahalanobis(A, true)
+Mahalanobis{Matrix{Int64}}([1 4 7; 2 5 8; 3 6 9])
+"""
 struct Mahalanobis{M<:AbstractMatrix} <: Metric
     qmat::M
+    Mahalanobis{M}(Q::M) where {M <: AbstractMatrix} = new{M}(Q)
+end
+function Mahalanobis(Q::M, skipchecks::Bool = false) where {M <: AbstractMatrix}
```
devmotion, line 27: The type parameter is not needed here.

src/mahalanobis.jl
dkarrasch: I'm fine with that. I moved the constructor to the inner part and made the argument a keyword argument.

dkarrasch:
Is this good to go? Is this kind of a bugfix? Technically breaking, but perhaps only breaking nonintended usage?

dkarrasch:
Gentle bump. Or is the absence of comments approval? 😄

src/mahalanobis.jl
```diff
@@ -1,11 +1,65 @@
...
+ERROR: ArgumentError: bilinear form is not symmetric/Hermitian
+
+julia> dist = Mahalanobis(A, skipchecks = true)
+Mahalanobis{Matrix{Int64}}([1 4 7; 2 5 8; 3 6 9])
+"""
 struct Mahalanobis{M<:AbstractMatrix} <: Metric
     qmat::M
+    function Mahalanobis(Q::AbstractMatrix; skipchecks::Bool = false)
+        if !skipchecks
+            ishermitian(Q) || throw(ArgumentError("bilinear form is not symmetric/Hermitian"))
```
nalimilan, line 27: Why not just say "matrix" instead of the more obscure "bilinear form"?

src/mahalanobis.jl
```diff
@@ -1,11 +1,65 @@
 # Mahalanobis distances
 
+"""
+    Mahalanobis(Q, skipchecks = false) <: Metric
```
nalimilan, line 4: I suspect we don't use spaces around `=` in general (but maybe worth checking). Also needs a `;`. Same below.
```suggestion
    Mahalanobis(Q; skipchecks=false) <: Metric
```

nalimilan (commented):
Just stylistic comments from my side, the API sounds OK to me (and I don't have an opinion on the math).

src/mahalanobis.jl
dkarrasch: Yeah, why not.

src/mahalanobis.jl
```diff
@@ -1,11 +1,65 @@
...
+
+# Example:
+```julia
+julia> A = collect(reshape(1:9, 3, 3)); Q = A'A;
+
+julia> dist = Mahalanobis(Q)
+Mahalanobis{Matrix{Int64}}([14 32 50; 32 77 122; 50 122 194])
+
+julia> dist = Mahalanobis(A)
+ERROR: ArgumentError: bilinear form is not symmetric/Hermitian
```
devmotion, line 18: Would be good to update the example I assume 🙂

src/mahalanobis.jl
```diff
@@ -1,11 +1,65 @@
...
+
+# Example:
+```julia
+julia> A = collect(reshape(1:9, 3, 3)); Q = A'A;
+
+julia> dist = SqMahalanobis(Q)
+SqMahalanobis{Matrix{Int64}}([14 32 50; 32 77 122; 50 122 194])
+
+julia> dist = SqMahalanobis(A)
+ERROR: ArgumentError: bilinear form is not symmetric/Hermitian
```
devmotion, line 49: Same here.

src/mahalanobis.jl
dkarrasch: 🤦

devmotion (commented):
I think the PR looks quite good. My main suggestion would be to make the symmetry check mandatory and to not include it in the `skipchecks` branch. This suggestion is motivated by the surprising behaviour observed in https://github.com/JuliaStats/Distances.jl/issues/174 and by the fact that this check is very cheap (basically zero cost) if users wrap the covariance matrix with `Symmetry` if they want to (I assume this is often done anyway to avoid numerical issues in other operations with the covariance matrix). The check for positive semi-definiteness is more expensive in general and hence it seems reasonable to provide an easy way to disable this check (even though this property is also assumed in the distance calculations silently to avoid negative distances caused by numerical issues for positive-semidefinite matrices).

nalimilan:
Something I'm realizing only now: throwing errors in cases that previously "worked" is breaking. There may be cases where matrices are approximately symmetric so results make sense, but that wouldn't pass the new checks, right? It would be safer to print warnings for now, and turn them into errors in the next breaking release.

test/test_dists.jl
```diff
@@ -405,6 +406,19 @@ end
             @test mahalanobis(x, y, Q) == sqrt(sqmahalanobis(x, y, Q))
             @test eltype(mahalanobis(x, y, Q)) == T
         end
+        A = rand(T, length(x), length(x))
+        S = A + A' - I
+        @test_throws DomainError SqMahalanobis(A)
```
dkarrasch, line 411: This test requires setting a seed. Maybe it's a bit too specific then? Also, the error is not thrown by Distances.jl, but by `eigmin` in `LinearAlgebra`.

nalimilan: You could hardcode a matrix that reproduces this? Otherwise we should use StableRNGs ideally.

dkarrasch: How about removing it for now, and add a `@test_throws` when we start to throw concrete errors from within `Distances.jl`? When we start to do that, we will hardly throw a `DomainError`, but an `ArgumentError`...

nalimilan: Why not, but you still want to check that the logs contain the warning, right?

dkarrasch: Yes!

dkarrasch:
Good to go now?

nalimilan (approved):

Issue #143, opened 2019-08-15
Quite different performance of pairwise on CPU vs GPU


xukai92:
I found the `pairwise` with `SqEuclidean` is faster than my own implementation on CPU but slower on GPU. Any idea why and possible optimization on Distances.jl side?

MWE:

```julia
using Test, BenchmarkTools, Distances, CuArrays

function pairwise_dot_kai(x)
    d, n = size(x)
    xixj = x' * x
    xsq = sum(x .^ 2; dims=1)
    return repeat(xsq, n, 1) + repeat(xsq', 1, n) - 2xixj
end

pairwise_dot(x) = pairwise(SqEuclidean(), x; dims=2)

xbench = randn(Float32, 784, 200);

@benchmark pairwise_dot_kai(xbench)
```
```
BenchmarkTools.Trial: 
  memory estimate:  1.52 MiB
  allocs estimate:  17
  --------------
  minimum time:     854.227 μs (0.00% GC)
  median time:      1.183 ms (0.00% GC)
  mean time:        1.361 ms (12.37% GC)
  maximum time:     125.259 ms (98.46% GC)
  --------------
  samples:          3662
  evals/sample:     1
```
```julia
@benchmark pairwise_dot(xbench)
```
```
BenchmarkTools.Trial: 
  memory estimate:  166.59 KiB
  allocs estimate:  204
  --------------
  minimum time:     359.751 μs (0.00% GC)
  median time:      406.615 μs (0.00% GC)
  mean time:        458.925 μs (6.46% GC)
  maximum time:     104.066 ms (99.27% GC)
  --------------
  samples:          10000
  evals/sample:     1
```
```julia
xbench = xbench |> cu;
@benchmark pairwise_dot_kai(xbench)
```
```
BenchmarkTools.Trial: 
  memory estimate:  1.20 MiB
  allocs estimate:  19424
  --------------
  minimum time:     19.042 ms (0.00% GC)
  median time:      20.028 ms (0.00% GC)
  mean time:        21.811 ms (3.62% GC)
  maximum time:     52.425 ms (38.23% GC)
  --------------
  samples:          230
  evals/sample:     1
```
```julia
@benchmark pairwise_dot(xbench)
```
```
BenchmarkTools.Trial: 
  memory estimate:  10.99 MiB
  allocs estimate:  240635
  --------------
  minimum time:     453.229 ms (0.00% GC)
  median time:      470.074 ms (0.00% GC)
[...]

PR #223, opened 2021-06-12
add Vectorization implementation for GPU

### LINKED ISSUE #137: how to calculate Euclidean distance using GPU ?
how to calculate Euclidean distance using GPU ?
I want to calculate Euclidean distance between 10000000*512 matrix and 10000000*512 matrix . 
Can I use  GPU ?? and how to do this??

johnnychen94:
One might even come up with a threaded `reduce` version for this, but that definitely belongs to future work.

```julia
@time using Distances
# PR: 0.284294 seconds (232.28 k allocations: 14.743 MiB, 30.94% compilation time)
# master: 0.051195 seconds (102.74 k allocations: 6.502 MiB, 85.42% compilation time)
using CUDA

X_cu = CUDA.randn(2^20);
Y_cu = CUDA.randn(2^20);

X = randn(2^20);
Y = randn(2^20);

@btime sqeuclidean($X, $Y)
# PR: 444.933 μs (0 allocations: 0 bytes)
# master: 418.690 μs (0 allocations: 0 bytes)
# almost a benchmark noisy

@btime sqeuclidean($X_cu, $Y_cu)
# GTX 3090
# PR: 51.827 μs (207 allocations: 5.64 KiB)
# master: take decades because of the scalar indexing
```

I did not expose this trait dispatching to the API level because I'm not sure if the user really wants this. Users might only want to control the computational device(e.g., GPU, CPU, CPU threads, CPU distributed) and may not want to control the under the hood implementation details.

---

> from @KristofferC in https://github.com/JuliaStats/Distances.jl/issues/143#issuecomment-522987543
> > Requires.jl makes loading this package 20x slower (#123 (comment)).

I still feel the runtime benefit worths it; not to mention that GPU support is basically unavailable before. Given that a lot of packages depend on `ArrayInterface`, adding this dependency won't add much of the loading time in any real project. (From [juliahub](https://juliahub.com/ui/Packages) results: `ArrayInterface` is depended by 692 packages while `Distances` is defended by 474 packages.)

```julia
julia> @time using ArrayInterface
  0.228708 seconds (224.97 k allocations: 13.884 MiB, 26.09% compilation time)

julia> @time using Distances
  0.014110 seconds (6.62 k allocations: 604.219 KiB)
```

Edit:

> From @nalimilan in https://github.com/JuliaStats/Distances.jl/issues/143#issuecomment-690931697
> > One solution to this would be to choose the best algorithm based on traits like those provided by ArrayInterface.

ArrayInterface requires Julia 1.2. I'll let you guys decide whether 1) we want to instead depend on Requires, or 2) directly bump the Julia lower bound here and drop Julia 1.0 and 1.1 support, or even 3) make a new package DistancesCUDA :(

closes #137

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,13 +3,15 @@ uuid = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 version = "0.10.3"
 
 [deps]
+ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
 [compat]
-julia = "1"
+ArrayInterface = "3.1.17"
 StatsAPI = "1"
+julia = "1"
 
 [extras]
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
--- a/Project.toml
+++ b/Project.toml
@@ -3,13 +3,11 @@ uuid = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
 version = "0.10.3"
 
 [deps]
-ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
 [compat]
-ArrayInterface = "3.1.17"
 StatsAPI = "1"
 julia = "1"
 
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -221,25 +221,26 @@ result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, p) where {Ta,Tb} =
     typeof(_evaluate(dist, oneunit(Ta), oneunit(Tb), oneunit(_eltype(p))))
 
 Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a, b, p=parameters(d))
-    _evaluate(infer_evaluate_strategy(d, a, b), d, a, b, p)
+    _evaluate(evaluate_strategy(d, a, b), d, a, b, p)
 end
 for M in (metrics..., weightedmetrics...)
     @eval @inline (dist::$M)(a, b) = _evaluate(dist, a, b)
 end
 
 # breaks the implementation into eval_start, eval_op, eval_reduce and eval_end
-function _evaluate(::Vectorization, d::UnionMetrics, a, b, ::Nothing)
+function _evaluate(::Broadcasting, d::UnionMetrics, a, b, ::Nothing)
     map_op(x,y) = eval_op(d, x, y)
     reduce_op(x, y) = eval_reduce(d, x, y)
     eval_end(d, reduce(reduce_op, map_op.(a, b); init=eval_start(d, a, b)))
 end
-function _evaluate(::Vectorization, d::UnionMetrics, a, b, p)
+function _evaluate(::Broadcasting, d::UnionMetrics, a, b, p)
     map_op(x,y,p) = eval_op(d, x, y, p)
     reduce_op(x, y) = eval_reduce(d, x, y)
     eval_end(d, reduce(reduce_op, map_op.(a, b, p); init=eval_start(d, a, b)))
 end
+_evaluate(::AbstractEvaluateStrategy, d::UnionMetrics, a, b, p) = error("Not implemented.")
 
-Base.@propagate_inbounds function _evaluate(::ScalarMapReduce, d::UnionMetrics, a, b, ::Nothing)
+Base.@propagate_inbounds function _evaluate(::MapReduce1, d::UnionMetrics, a, b, ::Nothing)
     @boundscheck if length(a) != length(b)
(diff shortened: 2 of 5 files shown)


Project.toml
```diff
@@ -3,13 +3,15 @@ uuid = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
...
 
 [deps]
+ArrayInterface = "4fba245c-0d91-5ea0-9b3e-6abc04ee57a9"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
 [compat]
-julia = "1"
+ArrayInterface = "3.1.17"
```
johnnychen94, line 12: `device(::CuArray)` was broken in earlier versions. https://github.com/JuliaArrays/ArrayInterface.jl/issues/165

src/metrics.jl
```diff
@@ -220,13 +220,26 @@ result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, ::Nothing) where {Ta,Tb}
 result_type(dist::UnionMetrics, ::Type{Ta}, ::Type{Tb}, p) where {Ta,Tb} =
     typeof(_evaluate(dist, oneunit(Ta), oneunit(Tb), oneunit(_eltype(p))))
 
-Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a, b)
-    _evaluate(d, a, b, parameters(d))
+Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a, b, p=parameters(d))
+    _evaluate(infer_evaluate_strategy(d, a, b), d, a, b, p)
+end
+for M in (metrics..., weightedmetrics...)
+    @eval @inline (dist::$M)(a, b) = _evaluate(dist, a, b)
```
johnnychen94, line 227: This is moved from L327-L329.

src/metrics.jl
```diff
@@ -373,7 +382,14 @@ totalvariation(a, b) = TotalVariation()(a, b)
...
 # if only NaN, will output NaN
-Base.@propagate_inbounds eval_start(::Chebyshev, a, b) = abs(first(a) - first(b))
+Base.@propagate_inbounds function eval_start(d::Chebyshev, a, b)
+    T = result_type(d, a, b)
+    if any(isnan, a) || any(isnan, b)
+        return convert(T, NaN)
+    else
+        zero(T) # lower bound of chebyshev distance
+    end
+end
```
johnnychen94, line 392: This is rewriting to support CuArray; scalar indexing `first` is slow for `CuArray`.

KristofferC:
A quite significant effort has been made to make adding this package as a dependency a no brainer by making it have few dependencies and thus a fast load time. Unfortunately, looking at the dependency here, a 10x load time increase is observed:


```
julia> @time using ArrayInterface
  0.184645 seconds (143.03 k allocations: 9.379 MiB, 2.25% compilation time)
```

```
julia> @time using Distances
  0.017180 seconds (18.09 k allocations: 1.566 MiB, 30.99% compilation time)
```

Some possible suggestions:

- Is it possible to profile `ArrayInterface` and make its load time in the same ballpark as Distances?
- Is it possible to implement GPU support using standard Julia constructs like broadcasting?
- Perhaps this code should go into one of the GPU libraries to not make everyone absorb this increased load time.

johnnychen94:
> Is it possible to profile ArrayInterface and make its load time in the same ballpark as Distances?

No, as far as I've understood. `ArrayInterface` heavily requires `Requires`

> Is it possible to implement GPU support using standard Julia constructs like broadcasting?

Yes, if you are referring to the vectorization version I just added here. But it's still slower in CPU due to the additional memory allocation for intermediate results.

> Perhaps this code should go into one of the GPU libraries to not make everyone absorb this increased load time.

The only place that requires `ArrayInterface` is `infer_evaluate_strategy`, we could only provide a fallback version here to unconditionally redirect to `ScalarMapReduce`, i.e., `infer_evaluate_strategy(d::PreMetric, a, b) = ScalarMapReduce()`. Then extends this method in other packages (e.g., DistancesCUDA.jl)

But I really don't think there's a need to be so conservative. For example, if we introduce LoopVectorization here and bring CPU runtime say >1.2x faster, does the overhead in loading time really matters? (Note that `LoopVectorization` depends on `ArrayInterface`)

KristofferC:
Looking a bit more at ArrayInterface it feels like it is way out of scope to add it to something so stable and lean as Distances.jl. For example, it does a lot of type piracy. I think it makes sense to add this to one of the packages that are in the "ArrayInterface" ecosystem (with possible modifications to Distance.jl to allow that to happen).

nalimilan:
It's too bad that ArrayInterface uses Requires.jl. I wish it would be a very lightweight package that others depend on, rather than the contrary. Or move the basic definitions to ArrayInterfaceBase so that packages can depend on that.

johnnychen94:
> Or move the basic definitions to ArrayInterfaceBase so that packages can depend on that.

I prefer to have something like `DistancesEx.jl` in JuliaStats that allows heavier dependency for speed and glue purposes. Possible efforts to bring LoopVectorization support can be made there as well.

Or maybe we can just work in the same Distances repo `/DistancesEx` as a sub-dir package like [SnoopCompile](https://github.com/timholy/SnoopCompile.jl) does.

devmotion:
If the main goal is GPU support, maybe one could also just create a GPU-specific DistancesCUDA package, similar to https://github.com/FluxML/NNlibCUDA.jl, with implementations of `_evaluate(::UnionMetrics, a::CUDA.CuArray, b::CUDA.CuArray)` etc.

johnnychen94:
Close it as ArrayInterface is too heavy a dependency for Distances. Without using `ArrayInterface`, it would be better to work out a `DistancesCUDA.jl` package with tweaked CUDA kernels.

PR #224, opened 2021-06-13
Pairwise and colwise with convenience functions

devmotion:
Currently, convenience functions such as `sqeuclidean` can't be used with `pairwise` and `colwise`:
```julia
julia> using Distances

julia> colwise(SqEuclidean(), rand(2), rand(2))
2-element Vector{Float64}:
 0.413932013463407
 0.11724482146295727

julia> colwise(sqeuclidean, rand(2), rand(2))
ERROR: MethodError: no method matching colwise(::typeof(sqeuclidean), ::Vector{Float64}, ::Vector{Float64})
Closest candidates are:
  colwise(::PreMetric, ::Any, ::Any) at /home/david/.julia/packages/Distances/gnt89/src/generic.jl:126
  colwise(::PreMetric, ::AbstractVector{T} where T, ::AbstractMatrix{T} where T) at /home/david/.julia/packages/Distances/gnt89/src/generic.jl:154
  colwise(::PreMetric, ::AbstractMatrix{T} where T, ::AbstractVector{T} where T) at /home/david/.julia/packages/Distances/gnt89/src/generic.jl:160
Stacktrace:
 [1] top-level scope
   @ REPL[9]:1

julia> pairwise(SqEuclidean(), rand(2), rand(2))
2×2 Matrix{Float64}:
 0.320931  0.689046
 0.115005  0.363253

julia> pairwise(sqeuclidean, rand(2), rand(2))
ERROR: MethodError: no method matching pairwise(::typeof(sqeuclidean), ::Vector{Float64}, ::Vector{Float64})
Closest candidates are:
  pairwise(::PreMetric, ::Any, ::Any) at /home/david/.julia/packages/Distances/gnt89/src/generic.jl:340
  pairwise(::PreMetric, ::Any) at /home/david/.julia/packages/Distances/gnt89/src/generic.jl:347
Stacktrace:
 [1] top-level scope
   @ REPL[6]:1
```

`pairwise(sqeuclidean, ...)` works after loading the fallback definition in StatsBase. Of course, `colwise` still fails and it does not use the implementation in Distances.
```julia
julia> using StatsBase

julia> pairwise(sqeuclidean, rand(2), rand(2))
2×2 Matrix{Float64}:
 0.00214758  0.0477589
 0.084275    0.213904
```

This PR fixes the problem for a large class of distances by defining `const sqeuclidean = SqEuclidean()` instead of `sqeuclidean(a, b) = SqEuclidean()(a, b)`: it ensures that `sqeuclidean isa SqEuclidean` and hence everything works as expected. Unfortunately, this approach does not work for distances that depend on additional parameters such as weighted metrics.


johnnychen94 (approved):
A good but perhaps breaking change. If there's no specific reason, I'd suggest a minor version bump for this.

Might also be good to leave a comment about the reason why some convenient functions (e.g., `sqeuclidean`) are instances while the others (e.g., `wsqeuclidean`) are normal functions. (Obvious at the scope of this PR but it is not so obvious when people are reading the entire codebase)

dkarrasch:
I don't think it's breaking. It's more like a new feature. Since we are below v1.0, a patch bump should be fine according to semver.

devmotion:
I agree, I considered this to be non-breaking. However, I gave it some additional thought and it seems it would "break" code if users defined method overloads for both e.g. `SqEuclidean` and `typeof(sqeuclidean)` - however, assuming that the users defined them consistently it would not be completely breaking but it would create a "method overwritten" warning.

johnnychen94:
That's exact the case in my mind when I said this is perhaps breaking. Two different downstream packages might override the functionality at two different name level.
Since a patch can be always submitted to the broken repo, if there was, so a patch bump is also okay to me.

devmotion:
So what's the decision here? Should it be considered to be breaking?

johnnychen94:
I just pointed out that this can potentially be an issue, but it's very likely that it is not an issue at all so a patch release is also good. Given that this is still at 0.x release a patch version bump can free us from a lot of CompatHelper notifications to the entire ecosystem.

nalimilan:
Yeah I'd tend to think it's not breaking so that tagging a patch release is OK.

nalimilan (approved):

Issue #216, opened 2021-04-23
Why single precision by default in Haversine?


juliohm:
https://github.com/JuliaStats/Distances.jl/blob/369f58632f76ea02922c894bf50162206b24a41f/src/haversine.jl#L14

Most lat/lon datasets are double precision so I don't understand why the single precision here. Can we double it?

nalimilan:
See https://github.com/JuliaStats/Distances.jl/pull/176#discussion_r541371211. The idea was that if you have `Float64` inputs, the result with be promoted to `Float64`, but if you have lower precision inputs they will keep using `Float32`. What's your use case?

juliohm:
lat/lon values are stored as Float64 in most GIS datasets. Computing anything in single precision is pretty rare because lat/lon values differ in the very last digits of the double precision representation. That is why I got confused with Float32.

This internal design decision could be made explicit in the docstring so that end users coming from GIS won't question it again. I did a quick benchmark to see if the promotion from single to double precision changed the performance, but apparently it doesn't.

juliohm:
If we talk about the most common use case of the Haversine, it will be in GIS applications. And in this case I would just set the default to Float64 to avoid confusion. People interested in Haversine for other exotic uses like GPU for example, could explicitly provide other type parameter.

nalimilan:
Do you have a concrete example where this could be confusing?

@mkborregaard @dkarrasch What do you think?

juliohm:
Just instantiating the distance will show the type parameter in the REPL and this is confusing for someone expecting double precision by default.

dkarrasch:
I believe that having a "minimal" type here is (at the risk of human confusion) the better choice. There is no lack of precision for that number, and for higher-precision floats it will be promoted anyway. To me personally, it feels a bit like the `I` in `LinearAlgebra`, whose default type is `Bool` for the same reason. It would be a different story if performance was suffering, though. Then we should perhaps take care of the most common use case.

juliohm:
Thank you, I think it makes sense given that there is no performance penalty.

Perhaps a note in the docstring would be enough.

nalimilan:
See https://github.com/JuliaStats/Distances.jl/pull/226#discussion_r679439667: it turns out we can just use an `Int` value as the default.

mkborregaard:
great!

PR #226, opened 2021-07-29
Small improvements to Haversine

torfjelde:
This PR
- Adds default constructor `Haversine{T}()`, allow type-specification without having to provide the radius.
- Makes `Haversine` use `Int` by default and promotes according to input.

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -11,7 +11,8 @@ The computed distance has the unit of the radius.
 struct Haversine{T<:Number} <: Metric
     radius::T
 end
-Haversine() = Haversine(Float32(6_371_000))
+Haversine{T}() where {T<:Number} = Haversine(T(6_371_000))
+Haversine() = Haversine{Int}()
 
 function (dist::Haversine)(x, y)
     length(x) == length(y) == 2 || haversine_error(dist)
@@ -26,10 +27,10 @@ function (dist::Haversine)(x, y)
     a = sind(Δφ/2)^2 + cosd(φ₁)*cosd(φ₂)*sind(Δλ/2)^2
 
     # distance on the sphere
-    2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
+    2 * (dist.radius * asin( min(√a, one(a)) )) # take care of floating point errors
 end
 
-haversine(x, y, radius::Number=Float32(6_371_000)) = Haversine(radius)(x, y)
+haversine(x, y, radius::Number=6_371_000) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 


src/haversine.jl
```diff
@@ -61,4 +62,4 @@ end
 
 spherical_angle(x, y) = SphericalAngle()(x, y)
 
-result_type(::Union{Haversine, SphericalAngle}, ::Type, ::Type) = Float64
```
KristofferC, line 64: Is the `SphericalAngle` `result_type` taken care of elsewher?

torfjelde: Oh crap, I didn't mean to remove the union; good catch :+1:

src/haversine.jl
```diff
@@ -61,4 +62,5 @@ end
...
+result_type(d::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} = promote_type(T1, T2, T3)
```
dkarrasch, line 65: I'm not sure this is correct. If all types are `<:Integer`, then the result is still some float because you're computing `[a]sin[d]`.

torfjelde: True! Make it `float(...)`?

Honestly wasn't aware of this `return_type` when I made the changes. Could we just use `Core.Compiler.return_type(d, map(typeof, (x, y)))` or is that a bad idea?

dkarrasch: I don't think there is anything wrong with the default `Float64` because that's what you will get from the `sin` computations anyway. If at some point somebody finds that `Float64` is too narrow for certain input types, then we can still debug that case.

torfjelde: > I don't think there is anything wrong with the default Float64 because that's what you will get from the sin computations anyway.

Maybe I misunderstand you but that's not the true, i.e. `sin(::Float32)::Float32` is true. This is exactly why I made this PR:) Even the current implementation is strange since by default `Haversine` uses `Float32` for the radius and so `(::Haversine)(x::Vector{Float32}, y::Vector{Float32})` returns `Float32` _but_ if you use it with `pairwise` it returns `eltype(...) = Float64` because of the `return_type` above.

torfjelde: Honestly, since the computations of the distances are seemingly "simple", you might be able to get away with using `Core.Compiler.return_type` to infer `result_type` for all distances instead of the hardcoding `result_type`.

dkarrasch: > so `(::Haversine)(x::Vector{Float32}, y::Vector{Float32})` returns `Float32`

True, I wasn't aware of that. Then yes, hardcoding `Float64` is not the right thing to do. Then I guess your `float` idea is the way to go. IIRC, then usage of `Core.Compile.return_type` is absolutely not recommended.

torfjelde: > Then I guess your float idea is the way to go.

Cool, I'll do that then :+1: 

> IIRC, then usage of Core.Compile.return_type is absolutely not recommended.

See I've heard the same, but IIRC the reason is that it might silently return `Any` but was thinking that since the functions used here are so "simple" it shouldn't be much of a problem. And in the cases where it is, then maybe `Any` is the best we can do?

But confirming/denying that hypothesis would require a bit more and so I don't think it's suitable for this PR anyways:)

src/haversine.jl
```diff
@@ -11,7 +11,8 @@ The computed distance has the unit of the radius.
 struct Haversine{T<:Number} <: Metric
     radius::T
 end
-Haversine() = Haversine(Float32(6_371_000))
+Haversine{T}() where {T<:Number} = Haversine(T(6_371_000))
+Haversine() = Haversine{Int}()
```
dkarrasch, line 15:
```suggestion
Haversine() = Haversine{Int32}()
```
? Otherwise there seems to be no chance for (in the default) case that `Float32` inputs yield `Float32` output.

src/haversine.jl
```diff
@@ -26,10 +27,10 @@ function (dist::Haversine)(x, y)
     a = sind(Δφ/2)^2 + cosd(φ₁)*cosd(φ₂)*sind(Δλ/2)^2
 
     # distance on the sphere
-    2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
+    2 * (dist.radius * asin( min(√a, one(a)) )) # take care of floating point errors
 end
 
-haversine(x, y, radius::Number=Float32(6_371_000)) = Haversine(radius)(x, y)
+haversine(x, y, radius::Number=6_371_000) = Haversine(radius)(x, y)
```
dkarrasch, line 33:
```suggestion
haversine(x, y, radius::Number=Int32(6_371_000)) = Haversine(radius)(x, y)
```

dkarrasch:
Looks good to me, besides the minor `Int32` suggestions. Could you add a few tests for the new functions, including things like checking the return type. I think that with this change, `Float32` input should yield a `Float32` output for the default radius.

src/haversine.jl
torfjelde: No it does:)

```julia
julia> promote_type(Int64, Float32)
Float32
```

dkarrasch: Excellent. I should stop guessing and test things on my own. :-))))) A test would be good then.

torfjelde:
Have a look now:)

src/haversine.jl
```diff
@@ -61,4 +64,4 @@ end
 
 spherical_angle(x, y) = SphericalAngle()(x, y)
 
-result_type(::Union{Haversine, SphericalAngle}, ::Type, ::Type) = Float64
+result_type(d::SphericalAngle, ::Type{T1}, ::Type{T2}) where {T1<:Number,T2<:Number} = promote_type(T1, T2)
```
torfjelde, line 67:
```suggestion
result_type(::SphericalAngle, ::Type{T1}, ::Type{T2}) where {T1<:Number,T2<:Number} = promote_type(T1, T2)
```

src/haversine.jl
```diff
@@ -26,13 +27,15 @@ function (dist::Haversine)(x, y)
...
-    2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
+    2 * (dist.radius * asin( min(√a, one(a)) )) # take care of floating point errors
 end
 
-haversine(x, y, radius::Number=Float32(6_371_000)) = Haversine(radius)(x, y)
+haversine(x, y, radius::Number=6_371_000) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 
+result_type(d::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} = promote_type(T1, T2, T3)
```
torfjelde, line 37:
```suggestion
result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} = promote_type(T1, T2, T3)
```

src/haversine.jl
dkarrasch: We still need the `float` of the promote_type, don't we?

dkarrasch: You could check with integer arguments.

torfjelde: Oh good catch; forgot about that! Doneso.

src/haversine.jl
```diff
@@ -26,13 +27,15 @@ function (dist::Haversine)(x, y)
...
-    2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
+    2 * (dist.radius * asin( min(√a, one(a)) )) # take care of floating point errors
 end
 
-haversine(x, y, radius::Number=Float32(6_371_000)) = Haversine(radius)(x, y)
+haversine(x, y, radius::Number=6_371_000) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 
+result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} = promote_type(T1, T2, T3)
```
torfjelde, line 37:
```suggestion
result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} = float(promote_type(T1, T2, T3))
```

src/haversine.jl
```diff
@@ -61,4 +64,4 @@ end
 
 spherical_angle(x, y) = SphericalAngle()(x, y)
 
-result_type(::Union{Haversine, SphericalAngle}, ::Type, ::Type) = Float64
+result_type(::SphericalAngle, ::Type{T1}, ::Type{T2}) where {T1<:Number,T2<:Number} = promote_type(T1, T2)
```
torfjelde, line 67:
```suggestion
result_type(::SphericalAngle, ::Type{T1}, ::Type{T2}) where {T1<:Number,T2<:Number} = float(promote_type(T1, T2))
```

torfjelde (commented):
Sorry, I'd completely forgotten about this PR! Added the `float` as discussed.

src/haversine.jl
nalimilan: It's really funny that we didn't think about it at https://github.com/JuliaStats/Distances.jl/issues/216. :-D

src/haversine.jl
```diff
@@ -26,13 +27,15 @@ function (dist::Haversine)(x, y)
...
-    2 * dist.radius * asin( min(√a, one(a)) ) # take care of floating point errors
+    2 * (dist.radius * asin( min(√a, one(a)) )) # take care of floating point errors
 end
 
-haversine(x, y, radius::Number=Float32(6_371_000)) = Haversine(radius)(x, y)
+haversine(x, y, radius::Number=6_371_000) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 
+result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} = float(promote_type(T1, T2, T3))
```
nalimilan, line 37: Please break lines at 92 chars.

torfjelde:
```suggestion
function result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number}
    float(promote_type(T1, T2, T3))
end
```

src/haversine.jl
```diff
@@ -26,13 +27,17 @@ function (dist::Haversine)(x, y)
...
 end
 
-haversine(x, y, radius::Number=Float32(6_371_000)) = Haversine(radius)(x, y)
+haversine(x, y, radius::Number=6_371_000) = Haversine(radius)(x, y)
 
 @noinline haversine_error(dist) = throw(ArgumentError("expected both inputs to have length 2 in $dist distance"))
 
+function result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number}
+    float(promote_type(T1, T2, T3))
+end
```
nalimilan, line 39: Actually, moving to the `function` syntax made the line wider than it needs to be. :-)
```suggestion
result_type(::Haversine{T1}, ::Type{T2}, ::Type{T3}) where {T1<:Number,T2<:Number,T3<:Number} =
    float(promote_type(T1, T2, T3))
```

dkarrasch (approved):

dkarrasch:
Good to go?

torfjelde:
Sorry, been AWOL for a couple of weeks here.

I believe so!

Issue #199, opened 2020-12-30
Define SemiMetric and Metric to be traits rather than types


matthieugomez:
The package `Distances` groups distances by their mathematical properties (symmetry and triangular inequality), via the type system `Metric <: SemiMetric <: Metric`.

This is limiting. In certain contexts, it makes more sense to group distances by their object of application (for instance string distances), or by their general algorithm (for instance, q-gram distances are a certain type of string distances that operate on qgrams). In `StringDistances`, I would love to be able to define certain functions only on these groups. Due to the lack of multiple inheritance in Julia, however, this is not possible.

For this reason, I'd prefer SemiMetric/Metric to be traits, rather than types. This would  keep the implementation of `pairwise`/`colwise` efficient, while making it easier for external packages to define their own abstract types.

nalimilan:
I'm afraid this would create dispatch issues, notably if we want `pairwise` to be supported in StatsBase for generic functions (https://github.com/JuliaStats/StatsBase.jl/pull/627), and specialized for distances in Distances. Or we would need to make `Metric` and friends internal types. But then what would you use traits for since you can't even dispatch on them?

Cc: @dkarrasch

dkarrasch:
It's hard for me to estimate the impact of this proposal without seeing a draft. In some sense, I think the proposal is consistent with @nalimilan's pairwise implementation in `StatsBase.jl`. There you have a `symmetric` keyword, which could be set by a trait `symmetric = issymmetric(dist)`, and in fact this is how I anticipated how one would map the generic `pairwise`s here to those in StatsBase. Currently, one would do that by setting `issymmetic(::SemiMetric) = true`, but without a type `SemiMetric`, one would do that one by one, similarly to how we handle `parameters` in `UnionMetric`s. Also, if I understand @matthieugomez correctly, then he suggests to replace only `Metric` and `SemiMetric` by traits, not `PreMetric`, so there would be still something that dispatches to `Distances.jl`. Did I understand that correctly?

PR #228, opened 2021-09-12
Remove Metric <: SemiMetric <: PreMetric

matthieugomez:
This PR removes the type structure `Metric, SemiMetric, PreMetric` which makes it harder to define new distance types (solving https://github.com/JuliaStats/Distances.jl/issues/199)

Still, generic distances are encouraged to define `issymmetric` and `issubadditive`, which is used in `pairwise` to accelerate computations.


KristofferC:
This doesn't look like an improvement to me and doesn't look like standard usage of traits. Traits are typically used for small pieces of extra information that you want to dispatch on (often related to dispatching to an optimized method). This might be the type of indexing (`LinearSlow` or `LinearFast`) or if the length is known (`HasLength`, `IsInfinite` etc). But not to encode big families of types like this.

From #199:

> In StringDistances, I would love to be able to define certain functions on these groups. This is not possible, however, due to the lack of multiple inheritance in Julia.

This feels like something that is more suitable for using traits for (which there should be no problem for you to do) so I don't really see that as an argument for this.

Also, remember this package is very old and has a large number of packages depending on it. Breaking changes should be kept to an absolute minimum.

matthieugomez:
Right it’s not even a trait. It ended up being simpler to define two
functions. I could not think of an example where one would need to dispatch
on premetric vs metric in a hot loop.

For StringDistances, traits won’t work since I would like to define
specialized methods for findall etc, which is not possible using traits.

I do get the argument about breaking changes, but the change is mainly
going to impact developers, rather than end users.

matthieugomez:
Another example is the following. Suppose I define a parametric type that corresponds to the normalized version of a distance (say, always between 0 and 1).
```
struct Normalized{T <: PreMetric}
   dist::T
```
There is currently no way to express that `Normalized(dist)` would by symmetric if `dist` is symmetric. With this PR, one could write
```
issymmetric(dist::Normalized) = issymmetric(dist.dist)
```

matthieugomez:
To me, the current situation of `Distances.jl` feels a little bit the same as if Base defined something like 
```julia
  abstract type SymmetricAbstractArray <: AbstractArray end
```
This would make it impossible create new array types (e.g. TriangularArray or SparseArray) that may end up being symmetric. 

- This exact problem is encountered within this package. At [some point](https://github.com/JuliaStats/Distances.jl/blob/a43d76b23874c10d8ab44909f7d4816ba88ef7db/src/metrics.jl#L200 ), the package tries to define a special distance type,  `UnionMetrics`. However, because of the forced inheritance `Metric <: SemiMetric <: PreMetric`, the package has to do the following workaround
   ```julia
   abstract type UnionPreMetric <: PreMetric
   abstract type UnionSemiMetric <: SemiMetric
   abstract type UnionMetric <: Metric
   const UnionMetrics = Union{UnionPreMetric, UnionSemiMetric, UnionMetric}
   ```
   This is inelegant, and also create some weird stuff --- in particular we don't have `UnionMetric <: UnionPreMetric`.

- This workaround does not always work. Suppose I define a parametric type that corresponds to the normalized version of a distance (say, always between 0 and 1).
  ```julia
  struct Normalized{T <: PreMetric} <: PreMetric
     dist::T
  end
   ```
   There is currently no way to express that `Normalized(dist)` remains symmetric if `dist` is symmetric. This is a big problem if I want to pass a normalized distance to a method that is only defined on `SemiMetric`s. In contrast, with this PR, one could simply write
  ```julia
  issymmetric(dist::Normalized) = issymmetric(dist.dist)
  ```

PR #232, opened 2021-10-26
Add specialized `pairwise` methods to `*msd`

dkarrasch:
I never used them, but just noticed that these metrics are defined in terms of other metrics that admit specialized `pairwise` implementations. So I figured these may benefit from that, and, no surprise, they do. At one point this allocates an extra vector, but on the other hand avoids recomputing `extrema` for each column of the first matrix.


devmotion (approved):
Looks good to me.

PR #235, opened 2021-11-28
Add efficient `SparseVector` method for some metrics

jlapeyre:
This PR makes the following efficient for `SparseVector`, partially addressing #5.

*  `bhattacharyya`, `bhattacharyya_coeff`
* `HellingerDist`, which follow directly from the previous item
*  non-weighted `UnionMetrics`. For example `euclidean(v1, v2)`. Specifically a method for `_evaluate(d::UnionMetrics, a::SparseVector, b::SparseVector, ::Nothing)` is included.

```julia
julia> n = 1000; dens = 0.1; v1 = sprand(n, dens); v2 = sprand(n, dens);

julia> @btime bhattacharyya($v1, $v2)  # before this PR
  14.877 μs (0 allocations: 0 bytes)
2.098185998707887

julia> @btime bhattacharyya($v1, $v2)  # after this PR
  240.100 ns (0 allocations: 0 bytes)
2.098185998707887

julia> n = 100; dens = 0.1; v1 = sprand(n, dens); v2 = sprand(n, dens);

julia> @btime bhattacharyya($v1, $v2) # before this PR
  714.130 ns (0 allocations: 0 bytes)
1.251741974341252

julia> @btime bhattacharyya($v1, $v2)  # after this PR
  45.974 ns (0 allocations: 0 bytes)
1.251741974341252
```

EDIT: The method  `_binary_map_reduce1` mentioned below has been removed. The implementation has been simplified.

* ~The new method calls a method `_binary_map_reduce1` which is also introduced in this PR. It is general enough to handle other similar use cases. I appended the `1` in (anticipation of several of these methods, in analogy to [these functions](https://github.com/JuliaLang/julia/blob/5c357e948d7f6ca8f4bff844da7a5f08567d386c/stdlib/SparseArrays/src/sparsevector.jl#L1240-L1245) in `SparseArrays` in the stdlib.  Perhaps removing the `1` in the name until it is needed would be a good idea.~

* ~The function `_binary_map_reduce1` works at least if `f(0, 0) == 0` and `op(v, 0) == v`, as well as some other cases noted in the code.~

* ~The numbered functions linked to above are for binary maps, not for binary mapreduce. The reasons for implementing variants (those with numbers appended) is different. In the former case, it is because an output vector is built and the functions are split according to which inputs can produces a zero.~

* ~Which functions like `_binary_map_reduce1` to implement, and where to put them is to be decided. It may make sense to decide on some generally useful binary mapreduce helper functions and put them in `SparseArrays`. In the short term, I put one of them here. They are not in the API so they can easily be moved to another repo later.~

I also tried this function
```julia
_bhattacharyya_coeff(x, y) = (sum(sqrt, x .* y), sum(x), sum(y))
```
This is also efficient for `SparseVector`  because `sum` is efficient for a single `SparseVector`. Yesterday, I thought I found some cases in which it was faster than using `_binary_map_reduce1`, but I can't find them now. However, it is often within 10 or 20 percent as fast. It is much less complex, which makes it attractive. Also, `sum` uses pairwise summation, which is more accurate in general. Has `Distances.jl` consciously opted for speed over correctness? I can understand making this choice for image processing.

See also https://github.com/JuliaLang/julia/issues/43248


jlapeyre:
In the original PR, two useless loops are executed in `_binary_map_reduce1`. I think it's better to use a more specialized mapreduce helper function in order to avoid this. The compiler has enough information to realize that repeated
`one_nonz_count += 1` can be replaced by a single addition. But, I doubt it does this (`@code_llvm` shows the loops and operations) so the loops likely run.

I'll simplify the function.

jlapeyre:
I added support for `UnionMetrics`. I tested most of the metrics at the cli for efficiency. I added a test in the test suite.

jlapeyre:
These tests just convert dense matrices to sparse. More tests are needed to exercise code paths for different numbers of non-zeros.

src/metrics.jl
```diff
@@ -308,6 +308,53 @@ Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::AbstractArray, b
     end
 end
 
+eval_op_a(d, ai, b) = eval_op(d, ai, zero(eltype(b)))
+eval_op_b(d, bi, a) = eval_op(d, zero(eltype(a)), bi)
+
+Base.@propagate_inbounds function _evaluate(d::UnionMetrics, a::SparseVector, b::SparseVector, ::Nothing)
```
dkarrasch, line 314: I wonder if you can relax this to some sort of `AbstractSparseVector`, which should ideally include column views into sparse matrices. Then `colwise` and `pairwise` should also benefit from this "out-of-the-box"™️.

jlapeyre: Wow. I already pushed something along these lines before reading this comment. It's not clear to me what interface `AbstractSparseVector` supports. But, `SparseColumnView` was easily supported. This is the vector you get from `@view m[:, i]` for a sparse matrix `m`.

dkarrasch: Awesome, the performance benefit is really significant. Your example is 100x faster for me! I think we will want to release this directly once it's merged, so you could bump the patch version in the Project.toml right away.

dkarrasch:
For nice code coverage, can you add a quick test with two sparse vectors of (i) different lengths and (ii) both of length 0? That should yield 100% diff coverage, and then we're ready to go IMO.

jlapeyre:
> different lengths
You mean different lengths so that an error is thrown, right ?

src/bhattacharyya.jl
```diff
@@ -50,6 +50,34 @@ end
...
+
+    ia = 1; ib = 1
+    eval_op = (a, b) -> sqrt(a * b)
+    s = zero(eval_op(zero(eltype(a)), zero(eltype(b))))
+    @inbounds while ia <= ma && ib <= mb
+        ja = anzind[ia]
+        jb = bnzind[ib]
+        if ja == jb
+            v = eval_op(anzval[ia], bnzval[ib])
+            s = v + s
```
dkarrasch, line 69: If `eval_op` is used only once, why not
```suggestion
            s += sqrt(anzval[ia] * bnzval[ib])
```
and remove the function definition above? I guess these are left-overs from your earlier binary_mapreduce1 approach?

src/bhattacharyya.jl
```diff
@@ -50,6 +50,34 @@ end
...
+    anzind = nonzeroinds(a)
+    bnzind = nonzeroinds(b)
+    anzval = nonzeros(a)
+    bnzval = nonzeros(b)
+    ma = nnz(a)
+    mb = nnz(b)
+
+    ia = 1; ib = 1
+    eval_op = (a, b) -> sqrt(a * b)
+    s = zero(eval_op(zero(eltype(a)), zero(eltype(b))))
```
dkarrasch, line 63: I overlooked the other occurence of `eval_op`, but maybe spell it out and make it consistent with the dense version above?
```suggestion
    s = zero(typeof(sqrt(oneunit(eltype(a))*oneunit(eltype(b)))))
```

src/bhattacharyya.jl
```diff
@@ -50,6 +50,34 @@ end
...
+            s = v + s
+            ia += 1; ib += 1
+        elseif ja < jb
+            ia += 1
+        else
+            ib += 1
+        end
+    end
+    # efficient method for sum for SparseVectorView is missing
+    return s, sum(anzval), sum(bnzval)
```
dkarrasch, line 78: Don't you want to accumulate `anzval` and `bnzval` in the loop above, similar to the dense version?

dkarrasch (commented):
Just a few (minor style) comments.

src/bhattacharyya.jl
jlapeyre: I made the decision consciously. But, both ways are fine.

src/bhattacharyya.jl
jlapeyre: I couldn't decide how to set the type of `s`. I suppose if the methods for `sqrt` and `*` for the given input types are type stable, then it  doesn't matter. But, maybe there is a reason to prefer 1 over 0.

dkarrasch: IIRC, then that had to do with units. Like you want the right output unit, including the squaring and taking the square root. It could be, though, that this unit-logic works just as fine with `zero`, I don't recall.

src/bhattacharyya.jl
jlapeyre: I thought about it:
* Accumulating in loops takes some thought and is error prone. For example, I would need to add two more loops to accumulate properly (compare to the last two loops in the similar code for `UnionMetrics`.
* The accumulation happens inside conditionals. On the other hand, with `sum` an entire packed array is summed. This allows optimizations, such as simd (But, I don't know if `sum` employs any of these for `Vector{Float64}` since it uses pairwise summation.)
* I did not benchmark the different possibilities.

dkarrasch: I see, I think your second point alone is good enough.

dkarrasch:
Shall we leave a comment about assumptions underlying the sparse `UnionMetric` implementation? AFAIU, the only assumption is that `eval_reduce(d, s, eval_op(d, 0, 0)) == s`, right? Because you jump over zero pairs.

jlapeyre:
> the only assumption is that `eval_reduce(d, s, eval_op(d, 0, 0)) == s`

Yes, I think this is correct.

PR #237, opened 2021-12-20
non rectangular periodic

GiggleLiu:
In lattices, the periodic boundary condition is non-rectangular in most cases. It can be an arbitrary parallelogon.
It would be nice if the periods can be a matrix.

I am not very familiar with the design pattern of this package, if someone feel the added code is not qualified according to your standard, please feel free to modify this PR. Thanks!

## Example
```julia
julia> using Distances

julia> p = PeriodicEuclidean([10 0.0; 10.0 10.0])
PeriodicEuclidean{Matrix{Float64}}([10.0 0.0; 10.0 10.0])

julia> evaluate(p, [0.1, 0.2], [10.1, 10.3])
0.1000000000000012
```

### DIFF
--- a/src/metrics.jl
+++ b/src/metrics.jl
@@ -394,6 +394,20 @@ weuclidean(a, b, w) = WeightedEuclidean(w)(a, b)
     s3 = min(s2, p - s2)
     abs2(s3)
 end
+
+function (pe::PeriodicEuclidean{<:AbstractMatrix})(a::AbstractVector, b::AbstractVector)
+    p = parameters(pe)
+    ndim = length(a)
+    @assert size(p,1) == size(p, 2) == length(b) == ndim  # check dimension
+    s = p \ (a - b)      # decompose to p-basis, a - b = sx x⃗ + sy y⃗ + ...
+    mindistance = Inf
+    s = mod.(s, 1)       # move to first "quadrant"
+    @inbounds for k = 0:1<<ndim-1  # 2^{# of dimensions} possible choices
+        loc = sum(i->view(p,:,i) .* (k>>(i-1) & 1 == 0 ? s[i] : s[i]-1), 1:ndim)
+        mindistance = min(norm(loc), mindistance)
+    end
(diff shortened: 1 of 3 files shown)


src/metrics.jl
```diff
@@ -394,6 +394,20 @@ weuclidean(a, b, w) = WeightedEuclidean(w)(a, b)
     s3 = min(s2, p - s2)
     abs2(s3)
 end
+
+function (pe::PeriodicEuclidean{<:AbstractMatrix})(a::AbstractVector, b::AbstractVector)
+    p = parameters(pe)
+    ndim = length(a)
+    @assert size(p,1) == size(p, 2) == length(b) == ndim  # check dimension
```
dkarrasch, line 401: More in line with the surrounding code (the multiple dispatch would work on the fact the parameter `p` is a matrix:
```suggestion
function _evaluate(dist::PeriodicEuclidean, a::AbstractVector, b::AbstractVector, p::AbstractMatrix)
    ndim = LinearAlgebra.checksquare(p)
    @boundscheck if length(a) != length(b)
        throw(DimensionMismatch("first array has length $(length(a)) which does not match the length of the second, $(length(b))."))
    end
    @boundscheck if length(a) != ndim
        throw(DimensionMismatch("arrays have length $(length(a)) but basis vectors have length $ndim."))
    end
```
And maybe move the function a little to not disrupt the sequence of `eval_*` functions?

GiggleLiu:
I notice that the distance computed in this way does not cover all cases (actually I found a counter-example). I need to do more research and will comment back later. Maybe this is still an open question

https://math.stackexchange.com/questions/1364594/distance-in-a-periodic-system

GiggleLiu:
Note:
The problem of finding the distance with periodic boundary condition is called the closest vector problem, it is known as NP-hard. However, for low-dimensional system, solving it regorously might be possible, check this paper:

https://arxiv.org/abs/1504.01995

Issue #238, opened 2022-02-01
Signature of `pairwise!` not consistent with StatsAPI and StatsBase


devmotion:
It seems Distances uses `pairwise!(out, metric, a[, b])` whereas the docstring in StatsAPI and the implementation in StatsBase use `pairwise!(f, dest, x[, y])`.

PR #139, opened 2019-08-06
move implementations into type overloading (aka. functor)

johnnychen94:
This PR is related to #22, one can easily rename `evaluate` later if we reach a consensus.

By moving implementations into functor, `evaluate` now is simply a really thin wrapper here; `evaluate` is limited only in `generic.jl` now (line 24). 

This reduces the possibility of meeting method ambiguity introduced by inconsistent type specification. For example, the implementation of `CorrDist` is cleaner.

Another change is the new usage `dist(x, y)`, e.g., `Euclidean()(x, y)`.

P.S.
[1] Vscode auto-formats some codes here and I commit it as well.
[2] For `ImageDistances.jl`, it will be much easier to extend to `Colorant` elements without getting lost in method ambiguity and dispatch orders 😄 

cc: @andreasnoack @KristofferC @ararslan

PR #239, opened 2022-02-01
Change signature of `pairwise!` and `colwise!`

devmotion:
This PR fixes https://github.com/JuliaStats/Distances.jl/issues/238 and deprecates `pairwise!(r, dist, a[, b])` in favour of `pairwise!(dist, r, a[, b])` which is consistent with StatsAPI and StatsBase.

It seemed natural to keep `colwise!` consistent with `pairwise!`, and hence I also deprecated `colwise!(r, dist, a, b)` in favour of `colwise!(dist, r, a, b)`.

### DIFF
--- a/test/test_dists.jl
+++ b/test/test_dists.jl
@@ -388,10 +388,10 @@ end # testset
     @test_throws DimensionMismatch sqmahalanobis(q, q, Q)
     mat23 = [0.3 0.2 0.0; 0.1 0.0 0.4]
     mat22 = [0.3 0.2; 0.1 0.4]
-    @test_throws DimensionMismatch colwise!(mat23, Euclidean(), mat23, mat23)
-    @test_throws DimensionMismatch colwise!(mat23, Euclidean(), mat23, q)
-    @test_throws DimensionMismatch colwise!(mat23, Euclidean(), mat23, mat22)
-    @test_throws DimensionMismatch colwise!(mat23, Bregman(x -> sqeuclidean(x, zero(x)), x -> 2*x), mat23, mat22)
+    @test_throws DimensionMismatch colwise!(Euclidean(), mat23, mat23, mat23)
+    @test_throws DimensionMismatch colwise!(Euclidean(), mat23, mat23, q)
+    @test_throws DimensionMismatch colwise!(Euclidean(), mat23, mat23, mat22)
+    @test_throws DimensionMismatch colwise!(Bregman(x -> sqeuclidean(x, zero(x)), x -> 2*x), mat23, mat23, mat22)
     @test_throws DimensionMismatch Bregman(x -> sqeuclidean(x, zero(x)), x -> 2*x)([1, 2, 3], [1, 2])
     @test_throws DimensionMismatch Bregman(x -> sqeuclidean(x, zero(x)), x -> [1, 2])([1, 2, 3], [1, 2, 3])
     sv1 = sprand(10, .2)
@@ -559,6 +559,7 @@ function test_colwise(dist, x, y, T)
         r2 = zeros(T, n)
         r3 = zeros(T, n)
         r4 = zeros(T, 1, n)
+        r5 = zeros(T, 1, n)
         for j = 1:n
             r1[j] = dist(x[:, j], y[:, j])
             r2[j] = dist(x[:, 1], y[:, j])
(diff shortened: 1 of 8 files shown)


devmotion:
Yes, this is definitely a strong argument against changing any of these function signatures. The main motivation for this PR is that [`pairwise!` is not owned by Distances anymore](https://github.com/JuliaStats/Distances.jl/pull/213) and that the [docstring in StatsAPI](https://github.com/JuliaStats/StatsAPI.jl/blob/00ce15f034e7ffdf16ec988766246755fcab47c4/src/StatsAPI.jl#L24) and the [implementation in StatsBase](https://github.com/JuliaStats/StatsBase.jl/blob/master/src/pairwise.jl) which owns `pairwise!` use the convention that the second argument is the one being mutated and the first one is the evaluated function. Since the [metrics are functors](https://github.com/JuliaStats/Distances.jl/pull/139) the different function signatures seemed inconsistent and surprising.

nalimilan:
It's true that `pairwise!` has lived in Distances for a long time before it was added to StatsBase and StatsAPI. But it makes sense to me to follow the convention used by `map!`, `broadcast!` and `get!` according to which the evaluated function comes first.

Given that there is no ambiguity, maybe we could allow both forms for some time? This is very similar to deprecating it, except that both would be mentioned in the docs.

test/test_dists.jl
```diff
@@ -673,8 +677,20 @@ function test_pairwise(dist, x, y, T)
...
+            rxy3 = similar(rxy)
+            @test pairwise!(dist, rxy3, vecx, vecy) ≈ rxy
+            @test rxy3 ≈ rxy
+
+            rxx2 = similar(rxx)
+            @test @test_deprecated(pairwise!(rxx2, dist, vecx)) ≈ rxx
+            @test rxx2 ≈ rxx
+            rxx3 = similar(rxx)
+            @test pairwise!(dist, rxx3, vecx) ≈ rxx
+            @test rxx3 ≈ rxx
```
nalimilan, line 693: Also test passing `dims`?

nalimilan (commented):
Sorry I completely missed this! Looks good, indeed the new order is consistent with `map!`. Just a small remark about a missing test, given that it's essential existing code continues to work for a long time all arguments should be tested.

dkarrasch:
Shall we finish this, release as v0.10.x, and subsequently merge #233 and release as v0.11.0? Potentially removing the deprecation warnings right away? Package maintainers will need to bump their `Project.toml` and would notice errors, whereas they will only get deprectation warnings as long as they don't touch their `Project.toml`.

test/test_dists.jl
devmotion: These tests here are only for general arguments such as iterators or vectors of vectors. There are actually no tests of `pairwise!` with `dims` currently. I added them in the last commit.

devmotion:
Sorry, I missed the last few comments here. I added more tests (it seems currently `pairwise!` with `dims` is not tested). I was not sure if the plan is to keep the deprecations in this PR, or if I should remove them as suggested above by @nalimilan.

nalimilan:
I'd say we can deprecate methods right away, as deprecations are not printed by default anyway so they won't annoy users in normal work. But as you prefer. What I wonder is how we should document these: should we attach a separate docstring to the methods, or add a note mentioning the deprecated argument order in the docstring for standard methods?

Also, how about tagging 0.10.8 with deprecated methods first, and then tag 1.0 so that we can drop deprecated methods in 1.x (e.g. in a few years as there's no hurry)?

devmotion:
@nalimilan I added docstrings to the deprecated methods.

(Regarding releases, I think maybe we might want to merge and release some other outstanding PRs such as #246 first before making a breaking release. I guess it will take a while for a breaking release to be supported throughout the Julia ecosystem and it might be nice to spread goodies such as #246 a bit faster.)

nalimilan (approved):
Looks good, but I'm a bit worried about the failing tests. They seem unrelated, but I couldn't find a way to run CI on master to confirm that.

devmotion:
I triggered CI in https://github.com/JuliaStats/Distances.jl/pull/250.

devmotion:
Indeed the same test failures show up in #250 (and hence on the master branch).

devmotion:
The weird thing is that I can't reproduce these test errors locally (Julia 1.9.1, clean environment). So I can't debug them locally.

@nalimilan Good to go anyway?

alyst:
This API change is appreciated, but it breaks definitions of custom distances.
It was a good reason to tag a new minor version.

devmotion:
The old syntax still exists and is only deprecated, so you can still call `colwise!` and `pairwise!` with the previous order of arguments. Hence it should not have been breaking. Do you have an example that broke?

alyst:
I've linked a PR with an example of `MySqEuclidean` metric that defined `pairwise!()` using old arguments order.
But the internal functions of `Distances.jl` were calling it, of course, with the new order.

devmotion:
This seems to be a bug, I should have updated the internal calls to `pairwise!` and `colwise!` only in a breaking release and left them unchanged until the deprecations are removed.

Issue #172, opened 2020-08-07
Distances + ChainRules


willtebbutt:
Are the maintainers of this package open to `ChainRules`-centric PRs? ie. are you happy to host "AD recipes" in this package?

The consensus is that the "correct" place for rules is in the package that defines the functions (otherwise it's type-piracy), in the same way that recipes for `RecipesBase` for plotting are hosted in the same package as the things that you want to plot using them.

An additional dependency on [ChainRulesCore](https://github.com/JuliaDiff/ChainRulesCore.jl) will be required, but as you can see from it's [Project.toml](https://github.com/JuliaDiff/ChainRulesCore.jl/blob/master/Project.toml), it is very lightweight.

KristofferC:
>  but as you can see from it's Project.toml, it is very lightweight.

Distances.jl is also very lightweight, for example

```

julia> @time using Distances
  0.039016 seconds (32.86 k allocations: 3.181 MiB)
```

but with ChainRulesCore as a dependency:

```
julia> @time using Distances
  0.092615 seconds (88.67 k allocations: 6.465 MiB)
```

willtebbutt:
Sure. So is that a tradeoff you're willing to make so that `Distances` can be made to work with AD?

PR #246, opened 2023-01-31
Add weak dependency on ChainRulesCore

devmotion:
This PR proposes to add a weak dependency on ChainRulesCore on Julia versions with support for extensions (i.e., Julia >= 1.9).

This would fix https://github.com/JuliaStats/Distances.jl/issues/172 (and https://github.com/FluxML/Zygote.jl/issues/1365) without affecting load and compilation times for users that do not load ChainRulesCore.

The initial set of reverse-mode rules in this PR covers the existing rules in Zygote and hence could replace these in Julia >= 1.9: Zygote could, e.g., wrap the Requires-block for Distances in a `if !isdefined(Base, :get_extension)` conditional.
The implementation in this PR is different from the code in https://github.com/FluxML/Zygote.jl/pull/923 since tests fail with the rules in the Zygote PR (e.g., due to type inference issues).
The rules in this PR are also all explicit and do not call back into the AD system, in contrast to the current implementation in Zygote.

IMO there are two main motivations for moving the rules to Distances.jl: It allows other ChainRules-compatible AD systems to use them as well without having to depend on Zygote and it fixes precompilation issues reported in https://github.com/FluxML/Zygote.jl/issues/1365. The second point could be fixed also by making Distances a weak dependency of Zygote but that would not solve the first problem.

I know that maintainers of Distances have tried to keep the package lightweight and to not add any dependencies that would increase load times. However, I think the [support for weak dependencies in Pkg](https://pkgdocs.julialang.org/dev/creating-packages/#Weak-dependencies) addresses these concerns. In a few other packages ChainRules-definitions were already moved to extensions recently, e.g., https://github.com/JuliaMath/ChangesOfVariables.jl/pull/12 and https://github.com/JuliaStats/LogExpFunctions.jl/pull/62.

### DIFF
--- a/ext/DistancesChainRulesCoreExt.jl
+++ b/ext/DistancesChainRulesCoreExt.jl
@@ -8,7 +8,12 @@ const CRC = ChainRulesCore
 
 ## SqEuclidean
 
-function CRC.rrule(::CRC.RuleConfig{>:CRC.HasReverseMode}, dist::SqEuclidean, x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
+function CRC.rrule(
+    ::CRC.RuleConfig{>:CRC.HasReverseMode},
+    dist::SqEuclidean,
+    x::AbstractVector{<:Real},
+    y::AbstractVector{<:Real}
+)
     Ω = dist(x, y)
 
     function SqEuclidean_pullback(ΔΩ)
--- a/ext/DistancesChainRulesCoreExt.jl
+++ b/ext/DistancesChainRulesCoreExt.jl
@@ -1,4 +1,4 @@
-module ChainRulesCoreExt
+module DistancesChainRulesCoreExt
 
 using Distances
 
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -11,3 +11,8 @@ using Unitful.DefaultSymbols
 
 include("F64.jl")
 include("test_dists.jl")
+
+# Test ChainRules definitions on Julia versions that support weak dependencies
+if isdefined(Base, :get_extension)
+    include("chainrules.jl")
+end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -13,6 +13,9 @@ include("F64.jl")
 include("test_dists.jl")
 
 # Test ChainRules definitions on Julia versions that support weak dependencies
-if isdefined(Base, :get_extension)
+# Support for extensions was added in
+# https://github.com/JuliaLang/julia/commit/93587d7c1015efcd4c5184e9c42684382f1f9ab2
+# https://github.com/JuliaLang/julia/pull/47695 
+if VERSION >= v"1.9.0-alpha1.18"
     include("chainrules.jl")
 end
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Distances"
 uuid = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
-version = "0.10.7"
+version = "0.10.8"
 
 [deps]
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
@@ -8,15 +8,24 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
 Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
+[weakdeps]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+
+[extensions]
+ChainRulesCoreExt = "ChainRulesCore"
+
 [compat]
+ChainRulesCore = "1"
 StatsAPI = "1"
 julia = "1"
 
 [extras]
+ChainRulesCore = "d360d2e6-b24c-11e9-a2a3-2a2ae2dbcce4"
+ChainRulesTestUtils = "cdddcdb0-9152-4a09-a978-84456f9df70a"
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 
 [targets]
-test = ["OffsetArrays", "Random", "Test", "Unitful"]
+test = ["ChainRulesCore", "ChainRulesTestUtils", "OffsetArrays", "Random", "Test", "Unitful"]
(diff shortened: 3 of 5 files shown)


KristofferC:
With extensions I think it is fine to add code like this even to light weight packages. I do think that https://github.com/JuliaStats/LogExpFunctions.jl/pull/64 should be taken into account.

simsurace:
Is there anything blocking this?

simsurace:
@KristofferC, could this be merged?

devmotion:
Bump :slightly_smiling_face:

nalimilan (commented):
Looks fine regarding the extension machinery AFAICT, but can somebody qualified review the rules themselves?

ext/DistancesChainRulesCoreExt.jl
```diff
@@ -0,0 +1,145 @@
+module DistancesChainRulesCoreExt
+
+using Distances
+
+import ChainRulesCore
+
+const CRC = ChainRulesCore
```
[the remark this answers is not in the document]
devmotion: I'm not sure, it was a while ago. I generally prefer explicit namespace declarations over import statements, but I assume I didn't want to write ChainRulesCore all the time and therefore added the alias (which I have seen in other packages and issues as well).

ext/DistancesChainRulesCoreExt.jl
```diff
@@ -0,0 +1,145 @@
...
+
+## SqEuclidean
+
+function CRC.rrule(::CRC.RuleConfig{>:CRC.HasReverseMode}, dist::SqEuclidean, x::AbstractVector{<:Real}, y::AbstractVector{<:Real})
```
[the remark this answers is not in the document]
devmotion:
```suggestion
function CRC.rrule(
    ::CRC.RuleConfig{>:CRC.HasReverseMode},
    dist::SqEuclidean,
    x::AbstractVector{<:Real},
    y::AbstractVector{<:Real}
)
```

devmotion:
> can somebody qualified review the rules themselves?

I wonder if maybe @sethaxen or @willtebbutt (who contributed to the rules in Zygote multiple times) would have time to check the rules?

willtebbutt (commented):
I think this broadly looks good.

Would you consider incorporating tests similar to those found here: https://github.com/FluxML/Zygote.jl/blob/29fa32a688fb4454d2ee81b9ba2a6484e4468bda/test/gradcheck.jl#L1212   ?

IIRC they came up in the context of some GP stuff, so it would be great to ensure that we don't regress.

simsurace:
Bump!

devmotion:
I'll return to the PR after my vacation, the PR is high up on my todo-list 🙂

devmotion:
I added tests with repeated columns: https://github.com/JuliaStats/Distances.jl/pull/246/commits/2e8563aa30dbb0568abc40ea4013a6f56cf4e49a Is this what you had in mind @willtebbutt?

devmotion:
Bump 🙂

willtebbutt (approved):
This is great -- sorry for the delay, this PR got lost in my inbox. Please merge when you're happy :)

devmotion:
Good to go, @nalimilan?

test/runtests.jl
```diff
@@ -11,3 +11,8 @@ using Unitful.DefaultSymbols
 
 include("F64.jl")
 include("test_dists.jl")
+
+# Test ChainRules definitions on Julia versions that support weak dependencies
+if isdefined(Base, :get_extension)
+    include("chainrules.jl")
+end
```
nalimilan, line 18: `Base.get_extension` is documented as internal so it could be dropped and we wouldn't notice that the test isn't run.
```suggestion
if VERSION >= v"1.10.0-DEV"
    include("chainrules.jl")
end
```

nalimilan (approved):

test/runtests.jl
devmotion: That seems wrong, it definitely exists in Julia 1.9.

While I can see the general point here, I think it's a rather hypothetical issue since the official way to detect whether weak dependencies are supported is this check (see, e.g., https://pkgdocs.julialang.org/v1/creating-packages/#Backwards-compatibility).

devmotion: BTW where is it documented as internal? In the docs (https://docs.julialang.org/en/v1/base/base/#Base.get_extension) it is not listed in the section "Internals" but "Code loading" it seems.

devmotion: Oh wow, I checked the version of the relevant commit (https://github.com/JuliaLang/julia/commit/495a004bda33e284b0acc612f5ced9ba1eb9a777), and it is `1.10.0-DEV.109` even though it exists in 1.9. Apparently, the feature was only backported to 1.9.

devmotion: But adding something like
```julia
if VERSION >= v"1.10.0-DEV.109"
    include("chainrules.jl")
end
```
would mean that these tests don't run on 1.9, wouldn't it? But we definitely want to test 1.9.

willtebbutt: 😬 do we know what the way forward with this is?

devmotion: Maybe we could change it to
```suggestion
# The VERSION check ensures that the tests are run even if
# Base.get_extension is removed
# Ref: https://github.com/JuliaStats/Distances.jl/pull/246#discussion_r1334654115
if VERSION >= v"1.10.0-DEV.109" || isdefined(Base, :get_extension)
    include("chainrules.jl")
end
```
@nalimilan, what do you think?

simsurace: I don't understand the issue if `isdefined(Base, :get_extension)` is the documented procedure to do this?

devmotion: Just making suggestions for how to resolve the discussions, I don't have a strong opinion regarding whether an additional check should be added or not 🙂 In any case, I think it just depends on what @nalimilan thinks.

simsurace: I appreciate the caution but this PR seems to be stuck for a long time for no good reason. @nalimilan already approved the PR, so the comment above is likely not a blocker for him.

devmotion: Given for how long the PR has been open and that @nalimilan was the first and main reviewer of the PR, I think there's no harm in giving him time to reply. I plan to merge end of next week if he hasn't disapproved explicitly by then.

nalimilan: Sorry. Yeah, I approved so that you can be free to merge as you like.

I think the commit on 1.9 is 327e0815d6dc7efef0cfaf82c2d2e9c01d55e304. `if VERSION >= v"1.9.0-alpha1.50"` should work on both 1.9 and 1.10.

devmotion: Ah, thanks for the pointer! Basing it on the backport commit is the cleanest way, of course, if we want to change it to a `VERSION` check. It seems the relevant backport commit that introduced `Base.get_extension` to 1.9 was actually https://github.com/JuliaLang/julia/commit/93587d7c1015efcd4c5184e9c42684382f1f9ab2 (the other one is "just" an improvement/bugfix for package extensions). In practice it won't matter but so I'll go with
```suggestion
# Support for extensions was added in
# https://github.com/JuliaLang/julia/commit/93587d7c1015efcd4c5184e9c42684382f1f9ab2
# https://github.com/JuliaLang/julia/pull/47695 
if VERSION >= v"1.9.0-alpha1.18"
    include("chainrules.jl")
end
```

PR #248, opened 2023-03-08
fix `_bhattacharyya_coeff` signature type

aviatesk:
After JuliaSparse/SparseArrays.jl/#355, `SparseArrays.SparseVectorUnion` is no longer `UnionAll`.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,6 +1,6 @@
 name = "Distances"
 uuid = "b4f34e82-e78d-54a5-968a-f98e89d6e8f7"
-version = "0.10.7"
+version = "0.10.8"
 
 [deps]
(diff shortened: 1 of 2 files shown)


aviatesk:
@dkarrasch Can you please review this and register a new version of Distances.jl? It is broken after JuliaSparse/SparseArrays.jl/#355 and now many packages depending on it is so too.

dkarrasch:
That should be only an issue on `nightly`, right?

dkarrasch:
Tests pass locally on latest nightly and on this PR.

dkarrasch:
Not sure what's going on on nightly, but the changes are safe. At the entry point, no check for numbers is ever made, so it's unclear why that restriction was added anyway.

PR #249, opened 2023-03-19
Reuse `spherical_angle` for `haversine`

navidcy:


dkarrasch:
I'm not sure this is a good idea. `sind`/`cosd` make quite some effort for accuracy. Did you test for differences? If we want to avoid code duplication, we could use a little meta-programming for that, à la:

```julia
for (d, sinfun, cosfun) in ((:Haversine, :sind, :cosd), (:SphericalAngle, :sin, :cos))
    @eval function (dist::$d)(x, y)
        length(x) == length(y) == 2 || haversine_error(dist)
        @inbounds λ₁, φ₁ = x
        @inbounds λ₂, φ₂ = y
        Δλ = λ₂ - λ₁  # longitudes
        Δφ = φ₂ - φ₁  # latitudes
        # haversine formula
        a = $sinfun(Δφ/2)^2 + $cosfun(φ₁)*$cosfun(φ₂)*$sinfun(Δλ/2)^2
        return (d === :SphericalAngle ? 2 * asin( min(√a, one(a)) ) : 2 * (dist.radius * asin( min(√a, one(a)) ))
    end
end
```
?

navidcy:
Nah... Let's leave it as is.

PR #250, opened 2023-07-01
DO NOT MERGE - Trigger CI

devmotion:
Just to rerun tests on master to see if the test failures in #239 are present on master as well.


devmotion:
As mentioned in https://github.com/JuliaStats/Distances.jl/pull/239#issuecomment-1618116911 I can't reproduce these errors locally - even though the package versions are the same.

nalimilan:
That's really weird. I can't reproduce it either. And the fact that it happens on all OSes seems to indicate it's not due to a particular machine or set up.

With some luck it will go away with Julia 1.9.2.

PR #251, opened 2023-07-02
Handle SparseArrays.jl dep as a Pkg extension

dkarrasch:
~~This is perhaps a slight abuse of the Pkg extension mechanism, and it is certainly breaking. Making the direct SparseArrays.jl dependency a weak dependency is a no-brainer. Unfortunately, this does not remove an indirect dependency on SparseArrays.jl through Statistics.jl. Now, strictly speaking we are not extending Statistics.jl, but it is only one metric that depends on it. Given that loading SparseArrays.jl on v1.10 is a rather heavy burden, avoiding it would be nice. Given that there was recent talking about releasing v1, and therefore an opportunity to break a few things, this is at least something to consider.~~

Unfortunately, the indirect strong SparseArrays.jl dependency remains via Statistics.jl.


dkarrasch:
I think we can make SparseArrays.jl a weak dependency of Statistics, the cut there is very clear. Then it suffices to make SparseArrays.jl a weak dependency here to avoid any, direct or indirect, dependency of Distances.jl on SparseArrays.jl. I'll make a PR there and then update this one here.

dkarrasch:
I assume this is now non-controversial and can be merged? In the current state we keep the direct dependency on Statistics.jl, hoping that it makes its own dependence on SparseArrays.jl weak (soon).

Project.toml
```diff
@@ -8,15 +8,24 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
...
 StatsAPI = "82ae8749-77ed-4fe6-ae5f-f523153014b0"
 
+[weakdeps]
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+
+[extensions]
+DistancesSparseArraysExt = "SparseArrays"
+
 [compat]
+Statistics = "0, 1"
```
devmotion, line 18: Can the Statistics-related changes be removed?
```suggestion
```

Project.toml
```diff
@@ -8,15 +8,24 @@ SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
...
 StatsAPI = "1"
 julia = "1"
 
 [extras]
 OffsetArrays = "6fe1bfb0-de20-5000-8ca7-80f57d26f881"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
+Statistics = "10745b16-79ce-11e8-11f9-7d13ad32a3b2"
```
devmotion, line 26:
```suggestion
```

dkarrasch:
Sure, those were left over.

devmotion (approved):
LGTM!

Everything should be fine since tests with nightly pass and the test failures on Julia 1.9 are known (and apparently caused by an LLVM bug: https://github.com/JuliaLang/julia/issues/50408), but maybe it would be even safer to merge #252 first and check that tests also pass on 1.9?

nalimilan:
Just to make sure, this isn't breaking, right?

dkarrasch:
No, it isn't. On older versions SparseArrays.jl remains a direct dependency, on newer versions it becomes a weak dependency. Packages/users that want the sparse optimizations that we move here must have SparseArrays loaded anyway.

nalimilan (approved):

dkarrasch:
Should I bump some version number? In StaticArrays.jl, we bumped the minor version, but that is already v1+.

nalimilan:
Just bump the patch version?

PR #258, opened 2023-11-08
Fix promotion of F64, add compat bounds for stdlibs

dkarrasch:
This is the recommended way to define promotion anyway, and this preempts failure related to https://github.com/JuliaLang/julia/pull/52038. Along the way, this also add compat bounds for updatable stdlibs as requested for future Julia versions.


devmotion:
Fixed in #260.

Issue #11, opened 2015-05-06
Distances with missing values


bdeonovic:
I saw this issue on DataFrames (https://github.com/JuliaStats/DataFrames.jl/issues/183), and noted that it was from a few years ago. Is there support for missing values in Distances.jl yet?

johnmyleswhite:
No, there is not. Pull requests would be reviewed I'm sure, although I imagine that Distances.jl shouldn't depend upon DataArrays.

lindahua:
I think supporting missing value is not in the scope of this package (at least at this point).

oxinabox:
We should come back and do this now that Missing data handling is in Base.

KristofferC:
There should be no need for special handling of Missing here.

Agreed with https://github.com/JuliaStats/Distances.jl/issues/11#issuecomment-100098293

rofinn:
FWIW, I found a specific use case where supporting `missing`s here would help the larger ecosystem. Specifically, Distances.jl -> NearestNeighbors.jl -> Impute.jl.

https://github.com/invenia/Impute.jl/issues/141#issuecomment-1870734968

Impute.jl isn't getting a lot of active development anymore, so I understand if it isn't worth the change, but updating Distances.jl and NearestNeighbors.jl is the best approach.

PR #261, opened 2023-12-29
RFC Added SkipMissing wrapper metric

rofinn:
A potential solution to https://github.com/JuliaStats/Distances.jl/issues/11

Pros:

- Logic is isolated to a few dozen lines in 1 file
- Independent of distance metric types
- Seems like a pretty common approach elsewhere (ex [scikit learn](https://scikit-learn.org/stable/modules/generated/sklearn.metrics.pairwise.nan_euclidean_distances.html))
- `skipmissing` already exists in base, so hopefully this approach is relatively intuitive

Cons:

- Performance costs (ie: generates a mask and views on each call)
- Sufficiently different from `Base.skipmissing` that we shouldn't export and could cause confusion?
- Arguably shouldn't be a metric and could probably just loosen other APIs to take a user defined function that calls `Missings.skipmissings`? (ex [NearestNeighbors.jl](https://github.com/KristofferC/NearestNeighbors.jl/blob/master/src/NearestNeighbors.jl#L26C81-L26C81))

### DIFF
--- a/src/missing.jl
+++ b/src/missing.jl
@@ -1,0 +1,47 @@
+"""
+Exclude any missing indices from being included the wrappped distance metric.
+"""
+struct SkipMissing{D<:PreMetric} <: PreMetric
+    d::D
+end
+
+result_type(dist::SkipMissing, a::Type, b::Type) = result_type(dist.d, a, b)
+
+# Always fallback to the internal metric behaviour
+(dist::SkipMissing)(a, b) = dist.d(a, b)
+
+# Special case vector arguments where we can mask out incomplete cases
+function (dist::SkipMissing)(a::AbstractVector, b::AbstractVector)
+    require_one_based_indexing(a)
+    require_one_based_indexing(b)
+    n = length(a)
+    length(b) == n || throw(DimensionMismatch("a and b have different lengths"))
+
+    mask = BitVector(undef, n)
+    @inbounds for i in 1:n
+        mask[i] = !(ismissing(a[i]) || ismissing(b[i]))
+    end
+
+    # Calling `_evaluate` allows us to also mask metric parameters like weights or periods
+    # I don't think this can be generalized to user defined metric types though without knowing
+    # what the parameters mean.
+    # NTOE: We call disallowmissings to avoid downstream type promotion issues.
+    if dist.d isa UnionMetrics
+        params = parameters(dist.d)
+
+        return _evaluate(
+            dist.d,
+            disallowmissing(view(a, mask)),
+            disallowmissing(view(b, mask)),
+            isnothing(params) ? params : view(params, mask),
+        )
+    else
+        return dist.d(
+            disallowmissing(view(a, mask)),
+            disallowmissing(view(b, mask)),
+        )
+    end
+end
+
+# Convenience function
+skipmissing(dist::PreMetric, args...) = SkipMissing(dist)(args...)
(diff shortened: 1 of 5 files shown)


src/missing.jl
```diff
@@ -0,0 +1,47 @@
+"""
+Exclude any missing indices from being included the wrappped distance metric.
+"""
+struct SkipMissing{D<:PreMetric} <: PreMetric
```
rofinn, line 4: I'm not sure if there's a benefit to have specific subtypes here? All the metrics I tested seemed to work.

src/missing.jl
```diff
@@ -0,0 +1,47 @@
...
+    @inbounds for i in 1:n
+        mask[i] = !(ismissing(a[i]) || ismissing(b[i]))
+    end
+
+    # Calling `_evaluate` allows us to also mask metric parameters like weights or periods
+    # I don't think this can be generalized to user defined metric types though without knowing
+    # what the parameters mean.
+    # NTOE: We call disallowmissings to avoid downstream type promotion issues.
+    if dist.d isa UnionMetrics
+        params = parameters(dist.d)
```
rofinn, line 30: AFAICT, `parameters` is only defined for `UnionMetrics`?

src/missing.jl
```diff
@@ -0,0 +1,47 @@
...
+
+        return _evaluate(
+            dist.d,
+            disallowmissing(view(a, mask)),
+            disallowmissing(view(b, mask)),
+            isnothing(params) ? params : view(params, mask),
+        )
+    else
+        return dist.d(
```
rofinn, line 39: I'm not sure if there's a safer fallback?

test/test_dists.jl
```diff
@@ -1051,6 +1051,124 @@ end
...
+            # Doesn't handle eltype Union{T, Missing}
+            @test_throws MethodError colwise(dist, A[:, 3:4], B[:, 3:4])
+        else
+            @test ismissing(dist(a, b))
+            @test_throws MethodError colwise(dist, A, B)
+            @test_throws MethodError pairwise(dist, A, B, dims=2)
+            @test_throws MethodError pairwise(dist, A, dims=2)
+        end
+
+        # Handle weights
```
rofinn, line 1154: We're just testing that `skipmissing` returns the same results as deleting indices missing in either vector.

test/test_dists.jl
```diff
@@ -1051,6 +1051,124 @@ end
...
+        if dist isa Distances.UnionMetrics && Distances.parameters(dist) isa Vector
+            @test D(a, b) == Distances._evaluate(dist, [1, 5], [6, 10], Distances.parameters(dist)[[1, 5]])
+        else
+            @test D(a, b) == dist([1, 5], [6, 10])
+        end
+
+        @test colwise(D, A, B)[3:4] ≈ colwise(dist, disallowmissing(A[:, 3:4]), disallowmissing(B[:, 3:4]))
```
rofinn, line 1161: Double check results are the same for non-missing columns.

test/test_dists.jl
```diff
@@ -1051,6 +1051,124 @@ end
...
+
+        M = pairwise(D, A, B, dims=2)
+        @test size(M) == (4, 4)
+        @test !any(ismissing, M)
+
+        M = pairwise(D, A, dims=2)
+        @test size(M) == (4, 4)
+        @test !any(ismissing, M)
```
rofinn, line 1169: Could manually check the calculations here, but I figured a few smoke tests that it doesn't error or return `missing`s was good enough?

rofinn:
Hmm, looks like some of the metrics throw different errors on different Julia versions.

Issue #151, opened 2019-10-09
SqPeriodicEuclidean?


SteffenPL:
For my simulation, the squared periodic euclidean distances has to be computed for many pairs.

Would it be possible to add `SqPeriodicEuclidean` to the collection of metrics or is it too special?
If wanted, I could work on an implementation and hand in a pull request.

dkarrasch:
Oooh, sorry, I overlooked the "Periodic". That's interesting! I didn't add it at the time, because I wasn't sure if anyone would ever use it. As a quick hack, you could call `pairwise(PeriodicEuclidean(periods), A, B); dims=2).^2`, or, indeed, make a PR. It would be nice if #150 got merged first, though. But otherwise, only few additions would be needed, because the modulo stuff is already there.

SteffenPL:
Cool. 

I'll wait for #150. Until then, I might figure out how to implement pairwise efficiently for a PR.

For now, I will use a user-defined metric as below in my project, since performance is not critical yet.

Thanks for your replies!

```julia
struct SqPeriodicEuclidean{W <: AbstractArray{<: Real}} <: SemiMetric
    periods::W
end

# PeriodicEuclidean
Base.eltype(d::SqPeriodicEuclidean) = eltype(d.periods)
@inline parameters(d::SqPeriodicEuclidean) = d.periods
@inline function eval_op(d::SqPeriodicEuclidean, ai, bi, p)
    s1 = abs(ai - bi)
    s2 = mod(s1, p)
    s3 = min(s2, p - s2)
    abs2(s3)
end
@inline function eval_op(d::SqPeriodicEuclidean, ai, bi)
    periods = d.periods
    p = isempty(periods) ? oneunit(eltype(periods)) : first(periods)
    eval_op(d, ai, bi, p)
end
@inline eval_reduce(::SqPeriodicEuclidean, s1, s2) = s1 + s2
@inline eval_end(::SqPeriodicEuclidean, s) = s

sqpeuclidean(a::AbstractArray, b::AbstractArray, p::AbstractArray{<: Real}) =
    SqPeriodicEuclidean(p)(a, b)
sqpeuclidean(a::Number, b::Number, p::Real) = SqPeriodicEuclidean([p])(a, b)


function Distances._evaluate(d::SqPeriodicEuclidean, a::AbstractArray, b::AbstractArray)
    @boundscheck if length(a) != length(b)
[...]

PR #265, opened 2024-10-09
Add `MinkowskiMetric` abstract type

eliascarv:
This new type will allow third-party packages to differentiate Minkowski metrics at the type level. It will also allow third-party packages to implement their own Minkowski metrics.

A practical example of a package that could use this new abstract type is NearestNeighbors.jl, which is currently using a union type for this purpose:

https://github.com/KristofferC/NearestNeighbors.jl/blob/master/src/kd_tree.jl#L1

With this new type NearestNeighbors.j could use Minkowski metrics defined in any package.

### DIFF
--- a/src/Distances.jl
+++ b/src/Distances.jl
@@ -9,6 +9,7 @@ export
     PreMetric,
     SemiMetric,
     Metric,
+    MinkowskiMetric,
 
     # generic functions
     result_type,
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -21,6 +21,13 @@ abstract type SemiMetric <: PreMetric end
 #
 abstract type Metric <: SemiMetric end
 
+# a minkowski metric is a metric that is defined by the formula:
+#
+#   d(x, y) = sum((x - y) .^ p) ^ (1 / p)
+#
+# where the `p` parameter defines the metric.
+abstract type MinkowskiMetric <: Metric end
+
 evaluate(dist::PreMetric, a, b) = dist(a, b)
 
 # Generic functions
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -22,7 +22,7 @@ abstract type SemiMetric <: PreMetric end
 abstract type Metric <: SemiMetric end
 
 """
-    MinkowskiMetric
+    MinkowskiMetric <: Metric
 
 A Minkowski metric is a metric that is defined by the formula:
 
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -21,13 +21,6 @@ abstract type SemiMetric <: PreMetric end
 #
 abstract type Metric <: SemiMetric end
 
-# a minkowski metric is a metric that is defined by the formula:
-#
-#   d(x, y) = sum((x - y) .^ p) ^ (1 / p)
-#
-# where the `p` parameter defines the metric.
-abstract type MinkowskiMetric <: Metric end
-
 evaluate(dist::PreMetric, a, b) = dist(a, b)
 
 # Generic functions
--- a/src/generic.jl
+++ b/src/generic.jl
@@ -21,6 +21,29 @@ abstract type SemiMetric <: PreMetric end
 #
 abstract type Metric <: SemiMetric end
 
+"""
+    MinkowskiMetric
+
+A Minkowski metric is a metric that is defined by the formula:
+
+`d(x, y) = sum(w .* (x - y) .^ p) ^ (1 / p)`
+
+where the `p` parameter defines the metric 
+and `w` is a potential weight vector (all 1's by default).
+
+Common Minkowski metrics:
+* `Cityblock`/`WeightedCityblock`: Minkowski metric with `p=1`;
+* `Euclidean`/`WeightedEuclidean`: Minkowski metric with `p=2`;
+* `Chebyshev`: Limit of `d` as `p` approaches infinity;
+* `Minkowski`/`WeightedMinkowski`: generic Minkowski metric for any `p`.
+
+## Notes
+* The difference between `Minkowski`/`WeightedMinkowski` and `MinkowskiMetric` 
+  is that while the first are generic Minkowski metrics, 
+  the second is the parent type of these metrics.
+"""
+abstract type MinkowskiMetric <: Metric end
+
 evaluate(dist::PreMetric, a, b) = dist(a, b)
 
 # Generic functions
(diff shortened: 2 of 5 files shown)


juliohm:
Appreciate if a maintainer can review and merge this.

juliohm:
ping @KristofferC @dkarrasch

src/metrics.jl
```diff
@@ -12,13 +12,15 @@ abstract type UnionSemiMetric <: SemiMetric end
 
 abstract type UnionMetric <: Metric end
 
+abstract type UnionMinkowskiMetric <: MinkowskiMetric end
+
```
dkarrasch, line 16: What do we need this for? Can't we replace any usage of `UnionMinkowskiMetric` by `MinkowskiMetric`? Users could simply subtype `MinkowskiMetric`, which is already abstract.

dkarrasch (commented):
This doesn't seem to hurt anyone, so this LGTM, except for the consideration of adding two types vertically into the type hierarchy.

src/metrics.jl
eliascarv: I created this type to follow the pattern of the other "Union" types:
```
abstract type UnionPreMetric <: PreMetric end

abstract type UnionSemiMetric <: SemiMetric end

abstract type UnionMetric <: Metric end
```

From what I understand, these "Union" types are treated differently in the implementation.

However, since all subtypes of `MinkowskiMetric` were subtypes of `UnionMetric`, it is possible to replace `UnionMinkowskiMetric` with `MinkowskiMetric` in the code.

dkarrasch: It's been a very long while when I was looking at/dealing with these things, but these types are needed for multiple dispatch AFAIR. They used to be just `Union`s, but then people couldn't reuse their infrastructure. Basically, the corresponding *metrics have the same computational pattern, just different operations in the details.

Couldn't you just subtype `MinkowskiMetric <: UnionMetric`? Thinking about it more, I think you actually should do that, because otherwise these types fall out of the `UnionMetric` type branch, which potentially other users rely on. You would then simply introduce another type layer, to distinguish some metrics (`MinkowskiMetric`) from other `UnionMetric` metrics.

eliascarv: So, is the PR code correct? Should we keep `UnionMinkowskiMetric`?

eliascarv: Done!

dkarrasch (commented):
Looks much cleaner now and non-breaking.

dkarrasch (approved):
LGTM

eliascarv:
@nalimilan, this PR is ready. Can you review it please?

src/metrics.jl
```diff
@@ -12,13 +12,21 @@ abstract type UnionSemiMetric <: SemiMetric end
 
 abstract type UnionMetric <: Metric end
 
+# a minkowski metric is a metric that is defined by the formula:
```
nalimilan, line 15: Make this a docstring (with appropriate formatting)? It would be useful to explain how this differs from `Minkowski` as it could confuse users. Could also be interesting to say what values of `p` subtypes correspond to.

Also:
```suggestion
# a Minkowski metric is a metric that is defined by the formula:
```

eliascarv: Done!

eliascarv: @nalimilan, do you think this new type should be exported?

nalimilan: Good question. In theory I would say yes, but the fact that `MinkowskiMetric <: UnionMetric` is problematic to me, as the latter is internal and its interface isn't documented (see https://github.com/JuliaStats/Distances.jl/issues/151#issuecomment-612460855 for how it apparently works).

Actually this raises a more fundamental point: in the current state of the PR, if people want to implement a `MinkowskiMetric` subtype, they have to implement the `UnionMetric` interface, or at least define enough methods so that none of the fallback methods defined for that type are used. This isn't logical. It would seem better to introduce `UnionMinkowskiMetric <: MinkowskiMetric <: PreMetric` and redefine `UnionMetrics` to include that (see https://github.com/JuliaStats/Distances.jl/pull/163 for context about `UnionMetric`). What do you think @dkarrasch?

eliascarv: @dkarrasch, the PR was like this at the beginning:

```julia
# /src/generic.jl
abstract type MinkowskiMetric <: Metric end

# /src/metrics.jl
abstract type UnionMinkowskiMetric <: MinkowskiMetric end
```

But after some discussions this was changed to the current code.

nalimilan: Ah sorry I had missed that discussion. This is definitely tricky. @dkarrasch You said:

> Couldn't you just subtype `MinkowskiMetric <: UnionMetric`? Thinking about it more, I think you actually should do that, because otherwise these types fall out of the `UnionMetric` type branch, which potentially other users rely on.

But why do you think it would be a problem that these types would not be `UnionMetric`? This is internal anyway. It seems to me that according to your reasoning, we could drop all `Union*` types, e.g. replacing `UnionMetric` with just `Metric`.

dkarrasch: After second thought, I think @nalimilan's proposal (which seems like the original state of the PR) is better. I confused `UnionMetric` with `UnionMetrics`, and what I was having in mind as the `UnionMetric` mechanism/infrastructure (the `eval_*` stuff) is actually a  `UnionMetrics` mechanism. Sorry for spreading confusion, and thanks @nalimilan for your care- and thoughtful review!

dkarrasch: @eliascarv Can you bump the patch version right in this PR? Then after we merge, we can release right away.

eliascarv: @nalimilan, so we can export the type `MinkowskiMetric` (not `UnionMinkowskiMetric`), right?

nalimilan: Yes I think so.

src/generic.jl
```diff
@@ -21,6 +21,29 @@ abstract type SemiMetric <: PreMetric end
 #
 abstract type Metric <: SemiMetric end
 
+"""
+    MinkowskiMetric
```
nalimilan, line 25:
```suggestion
    MinkowskiMetric <: Metric
```

src/Distances.jl
```diff
@@ -28,32 +29,24 @@ export
     Minkowski,
     Jaccard,
     BrayCurtis,
-    RogersTanimoto,
-
-    Hamming,
+    RogersTanimoto, Hamming,
```
nalimilan, line 32: Intended change?

eliascarv: No, it must have been the autoformat, undoing

dkarrasch (approved):

PR #266, opened 2024-10-29
Use `hypot` in haversine calculation for more floating-point safety

asinghvi17:
hypot(x, y) is sqrt(x^2 + y^2) but implemented in an error-corrected way.  It is a bit slower though, so depends on the tradeoff you want to make.  Feel free to accept or reject.

### DIFF
--- a/src/haversine.jl
+++ b/src/haversine.jl
@@ -24,10 +24,11 @@ function (dist::Haversine)(x, y)
     Δφ = φ₂ - φ₁  # latitudes
 
     # haversine formula
-    a = sind(Δφ/2)^2 + cosd(φ₁)*cosd(φ₂)*sind(Δλ/2)^2
+    # hypot(x, y) is equivalent to sqrt(x^2 + y^2)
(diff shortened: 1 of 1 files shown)


src/haversine.jl
```diff
@@ -24,10 +24,11 @@ function (dist::Haversine)(x, y)
     Δφ = φ₂ - φ₁  # latitudes
 
     # haversine formula
-    a = sind(Δφ/2)^2 + cosd(φ₁)*cosd(φ₂)*sind(Δλ/2)^2
+    # hypot(x, y) is equivalent to sqrt(x^2 + y^2)
+    sqrt_a = hypot(sind(Δφ/2), cosd(φ₁)*cosd(φ₂)*sind(Δλ/2))
```
dkarrasch, line 28: This is incorrect. The `cosd` coefficients are not squared, it's just the `sind` term.

asinghvi17: Ah whoops, I missed that somehow!  This is probably not very useful then :).

PR #270, opened 2025-03-27
Add link to TableDistances.jl

juliohm:
In  machine learning, distances are often computed with heterogeneous tables (e.g., kernel methods). This link can save users of Distances.jl a ton of time reinventing the wheel.


juliohm:
cc: @dkarrasch

